PR #12, opened 2020-03-20
Only set buffer-file-name locally in apheleia--write-region-silently

dsdshcym:
- Otherwise, buffer-file-name may be set to the buffer-file-name of
  the temporary file

Here is a screenshot of the changed buffer-file-name before this change:
image

### DIFF
--- a/apheleia.el
+++ b/apheleia.el
@@ -326,11 +326,9 @@ as in `write-region'. WRITE-REGION is used instead of the actual
   (funcall (or write-region #'write-region)
            start end filename append 0 lockname mustbenew)
   (when (or (eq visit t) (stringp visit))
-    (setq buffer-file-name (if (eq visit t)
-                               filename
-                             visit))
-    (set-visited-file-modtime)
-    (set-buffer-modified-p nil)))
+    (let ((buffer-file-name (if (eq visit t) filename visit)))
+      (set-visited-file-modtime)
+      (set-buffer-modified-p nil))))
 
 (defun apheleia--write-file-silently (&optional filename)
   "Write contents of current buffer into file FILENAME, silently.


raxod502:
Thank for the report! I have pushed what I think is a more correct fix. Let me know if it resolves your bug.

dsdshcym:
Thanks for the update!

But unfortunately, https://github.com/raxod502/apheleia/commit/284d1c48f84665beead5eb86d161fbdfe4e6e6b1 doesn't fix my issue.

1. The `buffer-file-name` variable is still changed to the tmp file's file-name. 
image
2. Emacs starts a endless loop like this (not sure if it's because I set `vc-follow-symlinks` to true):

    Followed link to /var/folders/m0/qgyn76495599mchnw44qlrnw0000gn/T/buffer-content-D4u9P2
    Followed link to /var/folders/m0/qgyn76495599mchnw44qlrnw0000gn/T/buffer-content-tJ0MeY
    Followed link to /var/folders/m0/qgyn76495599mchnw44qlrnw0000gn/T/buffer-content-9UNZ71
    Followed link to /var/folders/m0/qgyn76495599mchnw44qlrnw0000gn/T/buffer-content-w02f64
    Followed link to /var/folders/m0/qgyn76495599mchnw44qlrnw0000gn/T/buffer-content-7WRoHh
    Followed link to /var/folders/m0/qgyn76495599mchnw44qlrnw0000gn/T/buffer-content-i8ctpH
    Followed link to /var/folders/m0/qgyn76495599mchnw44qlrnw0000gn/T/buffer-content-B1Y27f
    Followed link to /var/folders/m0/qgyn76495599mchnw44qlrnw0000gn/T/buffer-content-SlPVoY
    Followed link to /var/folders/m0/qgyn76495599mchnw44qlrnw0000gn/T/buffer-content-B8K5QP
    Followed link to /var/folders/m0/qgyn76495599mchnw44qlrnw0000gn/T/buffer-content-GIsqLQ
    Followed link to /var/folders/m0/qgyn76495599mchnw44qlrnw0000gn/T/buffer-content-gc1b1V
    Followed link to /var/folders/m0/qgyn76495599mchnw44qlrnw0000gn/T/buffer-content-w7I82d
    Followed link to /var/folders/m0/qgyn76495599mchnw44qlrnw0000gn/T/buffer-content-4VUQhF
    Followed link to /var/folders/m0/qgyn76495599mchnw44qlrnw0000gn/T/buffer-content-Sw7CTe
    Followed link to /var/folders/m0/qgyn76495599mchnw44qlrnw0000gn/T/buffer-content-MCK3CX
    Followed link to /var/folders/m0/qgyn76495599mchnw44qlrnw0000gn/T/buffer-content-8UQlIP
    Followed link to /var/folders/m0/qgyn76495599mchnw44qlrnw0000gn/T/buffer-content-cIfvlR
    Followed link to /var/folders/m0/qgyn76495599mchnw44qlrnw0000gn/T/buffer-content-9qCN5R
    Followed link to /var/folders/m0/qgyn76495599mchnw44qlrnw0000gn/T/buffer-content-sLsCg2
    Followed link to /var/folders/m0/qgyn76495599mchnw44qlrnw0000gn/T/buffer-content-nn66hl
    Followed link to /var/folders/m0/qgyn76495599mchnw44qlrnw0000gn/T/buffer-content-C3a5wP
    Followed link to /var/folders/m0/qgyn76495599mchnw44qlrnw0000gn/T/buffer-content-ZJRLKs
    Followed link to /var/folders/m0/qgyn76495599mchnw44qlrnw0000gn/T/buffer-content-FfkyIz
    Followed link to /var/folders/m0/qgyn76495599mchnw44qlrnw0000gn/T/buffer-content-5O6x1P
    Followed link to /var/folders/m0/qgyn76495599mchnw44qlrnw0000gn/T/buffer-content-ZbL8Pb
    Followed link to /var/folders/m0/qgyn76495599mchnw44qlrnw0000gn/T/buffer-content-d10ZtH
    Followed link to /var/folders/m0/qgyn76495599mchnw44qlrnw0000gn/T/buffer-content-HKv4Ml
    Followed link to /var/folders/m0/qgyn76495599mchnw44qlrnw0000gn/T/buffer-content-p3sUia
    Followed link to /var/folders/m0/qgyn76495599mchnw44qlrnw0000gn/T/buffer-content-Rvj2l5
    Followed link to /var/folders/m0/qgyn76495599mchnw44qlrnw0000gn/T/buffer-content-Jx9H38
    Followed link to /var/folders/m0/qgyn76495599mchnw44qlrnw0000gn/T/buffer-content-KL5M5e
    Followed link to /var/folders/m0/qgyn76495599mchnw44qlrnw0000gn/T/buffer-content-LtLvIt
    Followed link to /var/folders/m0/qgyn76495599mchnw44qlrnw0000gn/T/buffer-content-gTHPvS
    Followed link to /var/folders/m0/qgyn76495599mchnw44qlrnw0000gn/T/buffer-content-HivCMn
    error in process sentinel: Quit [2 times]

Also, I notice this bug only happens when the formatted result is the same as the buffer content.
When the formatted result is different than the buffer content, this issue doesn't happen.

Sorry for not being able to provide a reproducible steps. Maybe there's a deeper issue here but I can't find out.
Another thing I know is that this PR does fix the issue for me.

Anything I can help to debug this issue?

raxod502:
Let me know if the latest commit fixes it for you.

raxod502:
Oops. I was able to reproduce the behavior (unintentionally), and discovered it wasn't fixed. Try this commit instead.

dsdshcym:
https://github.com/raxod502/apheleia/commit/0c0085f246bcf9ff4beb9379c7012d32212870a7 still doesn't fix the issue for me.

Let me see if I can have a minimal Emacs config to reproduce this issue for you this weekend.

dsdshcym:
Sorry for the late reply.

I've identified the root cause today.

I'm using https://github.com/dgutov/diff-hl, specifically, `(diff-hl-flydiff-mode)`

Basically, `diff-hl-flydiff-mode` would create temporary file and cause this infinite loop when combined with `apheleia` format on save.

I've been using the branch in this PR for several weeks and the infinite loop never occurs.

I'm not sure why this PR fixes the issue, though. And I don't know if this issue is worth fixing. Anything I can do to help?

raxod502:
This PR is definitely not correct. It breaks the interface of `write-region`. Looking briefly at diff-hl-flydiff-mode, I would guess that the problem is it sets an idle timer to write a file after each time a file is written. To solve the bug, someone must determine the control flow of this code (how Apheleia and diff-hl-flydiff end up in a loop) and then make a proposal about what the right way to avoid it is.

dsdshcym:
Got it.
It's easier for me to just turn off `diff-hl-flydiff-mode` than making a proposal :joy:
So I'll close this PR for now.

Another thing we can do is to warn apheleia users about this issue in the README. I can open a PR if you think so, too. (Adding it to User Guide or opening a new section?)

raxod502:
I would add a Caveats section, much like [in the Selectrum README](https://github.com/raxod502/selectrum#caveats), mentioning that there is a conflict and describing what the behavior looks like.

PR #17, opened 2020-06-15
Try again to preserve undo history

### LINKED ISSUE #15: Seems to wreck undo history
When a lot of formatting is going on I seem to lose the ability to undo back very far. This is extremely bad for UX. Probably a consequence of `undo-limit` and the reformatting generating large volumes of undo data (not sure why).

leungbk:
Closes #15.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -38,9 +38,14 @@ The format is based on [Keep a Changelog].
   buffer that had no backing file. This has been fixed, and Apheleia
   can run correctly on buffers both with and without backing files
   ([#9]).
+* Under some circumstances Apheleia would cause the buffer's undo
+  history to be lost. An attempt has been made to fix this, although
+  it's not obvious how to reproduce the problem so the fix has not
+  been verified ([#17]).
 
 [#4]: https://github.com/raxod502/apheleia/issues/4
 [#9]: https://github.com/raxod502/apheleia/pull/9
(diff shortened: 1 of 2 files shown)


leungbk:
It seems that `undo-amalgamate-change-group` is not available prior to Emacs 26. Should we just paste in the source? Or should we simply drop support for 25?

raxod502:
Emacs 25 is still used enough that I think I'd prefer to continue supporting it. In particular it's shipped with the second-to-latest Ubuntu LTS which most people are probably still on who run Ubuntu LTS.

Pasting in the definition is fine, [here is an example of how to do that](https://github.com/raxod502/straight.el/blob/a7f94876b2bf96d2595706270be6630ecc94f0d3/straight.el#L88-L93).

leungbk:
I closed the PR since it didn't seem to do anything to alleviate the problem of large entries in `buffer-undo-list`.

Tha markers generated within `apheleia--apply-rcs-patch` seem to be accounting for much of the size of each entry within `buffer-undo-list`. But from a glance at the code, the markers look pretty important, so I'm not sure how the space usage of this function can be cut down.

raxod502:
Oh, huh. I didn't realize markers got recorded in the undo history. I guess that makes sense though. Do we clean them up properly? I don't see anything for that in the code. Perhaps they are accumulating over time and the undo history is getting exponentially bigger because of all the markers getting moved around.

PR #24, opened 2021-01-16
[DRAFT] More formatters

PrimaryCanary:
I'll be adding many more formatters in the next few days. I hope to reach parity with [`emacs-format-all-the-code`](https://github.com/lassik/emacs-format-all-the-code/). Each one will be a separate commit for easy cherry-picking. 

I've read your contributor guide. I'll update the changelog and such when I'm closer to finishing this pull.


raxod502:
Wow, this is phenomenal. Thanks!

raxod502:
I hope you don't mind---I've updated the changelog, merged in master, and included this for the next release. We can always add even more formatters in the future!

adimit:
Hi, that's somewhat bad timing right now, because I've just been writing an issue about integrating `rustfmt`. I'm afraid the proposed PR will not work for `rustfmt`. I'll post the issue in a moment.

Issue #2, opened 2019-07-18
Point is sometimes moved small distances


raxod502:
Example:

```
  if (node.type === "CallExpression" && (node.callee.type === "Import" || (node.callee.type === "Identifier" && node.callee.name === "require"))) {
    //@
  }
```

with point represented by `@` is reformatted by Prettier to:

```
  if (
    node.type === "CallExpression" &&
    (node.callee.type === "Import" ||
      (node.callee.type === "Identifier" && node.callee.name === "require"))
  ) {@
    //
  }
```

instead of:

```
  if (
    node.type === "CallExpression" &&
    (node.callee.type === "Import" ||
      (node.callee.type === "Identifier" && node.callee.name === "require"))
  ) {
    //@
  }
```

I suspect that this problem could be solved by tweaking the weights used in the dynamic programming algorithm.

jpablobr:
I was having that pointer moving out of place problem and seems like I was able to fix it by setting  the `--line-length` in apheleia to 78. What's weird is that the only thing I have I have using 78 is the `warning-fill-column` variable which doesn't seem like it can be related, so not sure.

```elisp
  (setf (alist-get 'black apheleia-formatters)
      '("black" "--line-length" "78" "-"))
```

But does seem like Black/Emacs format-conflicts can lead to these types of issues.

raxod502:
I think this is unlikely to be directly related to the line length configuration. I think what you're observing is just that the diff being applied is slightly different, which happens to trigger the (somewhat nondeterministic) limitation in Apheleia's dynamic programming algorithm that is the cause of point moving.

raxod502:
This is fixed now #290 is merged.

PR #26, opened 2021-01-19
Basic Automated Test Harness

PrimaryCanary:
Here is a very simple framework for ensuring all the formatters properly configured. Currently, it has a small issue in that it doesn't actually work.

Each test consists of two files of the format `formatters/mode-it-should-be-active-in/formatter.ext` and `formatters/mode-it-should-be-active-in/formatter.ext.formatted`. The former is unformatted code while the latter is the expected output after running the formatter. `test-apheleia.el` should (but doesn't) return the result of calling the given formatter on the unformatted code to stdout.

All output from `test-apheleia.el` should be coming from the `apheleia--run-formatters` function. I think any stdout is being smothered because it's called from the async inferior process. How would I make `test-apheleia.el` return the formatted buffer to stdout?

### TODO:
- [x] Make `test-apheleia.el` work.
- [x] Ensure all tests contain the proper text.
- [x] Add supported modes and formatters table to README.
- [x] Add test to ensure README table matches `apheleia-modes-alist` and `apheleia-formatters`.
- [x] Run tests from Docker container.
- [x] Add a test target in the Makefile.
- [x] Add [package-lint](https://github.com/purcell/package-lint) target in the Makefile.
- [x] Run the tests in CI.


scripts/docker-install.bash
```diff
@@ -10,8 +10,12 @@ fi
 
 uid="$1"
 
-packages="
+# Hashicorp repo for Terraform
+curl -fsSL https://apt.releases.hashicorp.com/gpg | gpg --dearmor | \
+    sudo tee /usr/share/keyrings/docker-ce-archive-keyring.gpg > /dev/null
+sudo apt-add-repository -y "deb [arch=amd64] https://apt.releases.hashicorp.com $(lsb_release -cs) main"
```
raxod502, line 16: I don't think sudo is necessary. We should be running as root here.

scripts/docker-install.bash
```diff
@@ -24,11 +28,53 @@ git
...
 export DEBIAN_FRONTEND=noninteractive
 apt-get update
-apt-get install -y $(grep -v "^#" <<< "$packages")
+apt-get install -y $(grep -v "^#" <<< "$apt_packages")
+opam install $(grep -v "^#" <<< "$opam_packages")
+npm install --global $(grep -v "^#" <<< "$npm_packages")
+pip install $(grep -v "^#" <<< "$pip_packages")
+# Stack
+curl -sSL https://get.haskellstack.org/ | sh
+stack install $(grep -v "^#" <<< "$stack_packages")
```
raxod502, line 75: You may find that you are unsatisfied with Stack-installed packages. They must be built from source and often take 15+ minutes to install, which is suboptimal for CI. Many Haskell projects provide binary downloads, which are often a better option.

If, despite our best efforts, we can't manage to make the Docker installation take a reasonable amount of time, I can suggest ways that it could be optimized. (My work on https://github.com/raxod502/riju gives me some context on dealing with installing poorly packaged software into Docker images.)

test/formatters/web-mode/prettier-babel.html.formatted
```diff
@@ -0,0 +1,112 @@
...
+                           What level of logs to report.
+                           Defaults to log.
+  --require-pragma         Require either '@prettier' or '@format' to be present in the file's first docblock comment
+                           in order for it to be formatted.
+                           Defaults to false.
+  --stdin-filepath <path>  Path to the file to pretend that stdin comes from.
+  --support-info           Print support information as JSON.
+  -v, --version            Print Prettier version.
+
+
```
raxod502, line 112: Hmmm, this seems not quite right! :P

test/test-apheleia.el
```diff
@@ -0,0 +1,68 @@
+#!/usr/bin/env bash
+":"; exec emacs --script "$0" -- "$@" # -*- mode: emacs-lisp; lexical-binding: t; -*-
```
raxod502, line 2: Nice hack---I'm wondering what the `":"` is needed for, though?

test/test-formatters.bash
```diff
@@ -0,0 +1,11 @@
+#!/usr/bin/env bash
+set -euo pipefail
+
+find formatters/ -type f,l ! -name "*.formatted" | while read f;
```
raxod502, line 4: > `-type f,l`

Do we want to find symlinks, or just files?

test/test-apheleia.el
```diff
@@ -0,0 +1,68 @@
...
+       (hoist t))
+  (with-temp-buffer
+    (insert-file-contents f)
+    (apheleia--run-formatter
+     (alist-get (alist-get mode apheleia-mode-alist)
+                apheleia-formatters)
+     (lambda (buf)
+       (with-current-buffer buf
+         ;(buffer-string)
+         (write-file "/tmp/klsdfkljd"))))))
```
raxod502, line 68: > All output from test-apheleia.el should be coming from the apheleia--run-formatters function. I think any stdout is being smothered because it's called from the async inferior process. How would I make test-apheleia.el return the formatted buffer to stdout?

Do you just want to

```elisp
(princ (buffer-string))
```

? This should print to stdout.

raxod502 (commented):
I love how simple it is to add a test, just adding the two files and no extra busywork required!

scripts/docker-install.bash
PrimaryCanary: Fixed with a nasty curl one-liner.

test/formatters/web-mode/prettier-babel.html.formatted
PrimaryCanary: And that's why "Ensure all tests contain the proper text." is on the checklist :P

test/test-formatters.bash
PrimaryCanary: I've used symlinks for modes that are likely to be working on the same code with the same formatter e.g. rust-mode and rustic-mode both work on Rust and likely use rustfmt. I can strengthen the requirement if you like.

test/test-apheleia.el
PrimaryCanary: Just printing isn't quite enough. See https://github.com/raxod502/apheleia/pull/26/commits/749609076456402bae4271f63a7ef73e591463c4 for details.

test/test-apheleia.el
PrimaryCanary: Wonderful little monstrosity innit? That's how you do multiline shebangs with elisp.
https://rosettacode.org/wiki/Multiline_shebang#Emacs_Lisp

test/test-apheleia.el
raxod502: Oh, I see. The problem is your script was exiting before the asynchronous call finished. Yes, `accept-process-output` would block until it was finished.

PrimaryCanary:
The good:
- The Docker container successfully builds and runs all formatters.
- The tests will run in CI.
The bad:
- There are currently two formatters building from source. I couldn't find a binary for ocamlformat and the Brittany binary Didn't Work On My Machine<sup>TM</sup>.
- The container builds very slowly as a result. We might have to use your expertise to speed up CI and builds.
- No point tests which would make testing and fixing https://github.com/raxod502/apheleia/issues/2 easier.
The ugly:
- I haven't tested the CI settings because I don't use CircleCI. It looks like it will work.

If none of those issues are too big, I'd say this is ready to merge.

test/test-formatters.bash
raxod502: Fair enough, that makes sense.

raxod502:
I fixed things up so CI will run in only 2-3 minutes, and made various other improvements. The main thing remaining is that there seems to be something wrong with json-mode in Emacs 25, do you think you could take a look at that?

raxod502:
This thread is being closed automatically by [Tidier](https://github.com/raxod502/tidier) because it is labeled with "waiting on response" and has not seen any activity for 90 days. But don't worry—if you have any information that might advance the discussion, leave a comment and I will be happy to reopen the thread :)

raxod502:
I wrote my own version of this in https://github.com/raxod502/apheleia/pull/72, partly drawing from the code here. The only things missing are package-lint integration, and the auto-generated table of formatters in the README.

Issue #36, opened 2021-03-08
Don't change default-directory?


apeschar:
What's the reason that default-directory is set to the project root instead of left unchanged as the directory that the file is in?  If the project contains different formatter configs in different subfolders, the wrong one (or none) will get used.

In my `init.el`, I added:

~~~
(defun apheleia--project-root () default-directory)
~~~

But I'm wondering whether this functionality is not better removed altogether? Most if not all formatters should automatically find the right config in parent folders.

lassik:
Running most formatters in the file's director (not the project's) is the right thing.

Only a handful of formatters require a hack to run from the project root directory. Format-all's code shows which ones.

apeschar:
It looks like none of the formatters for which format-all changes default-directory are supported by prettier.  So until that is the case, I believe this behavior can safely be removed.

I will submit a PR.

raxod502:
Reverted this change in https://github.com/raxod502/apheleia/commit/87115f5cda2a6c0653fc0ac69bbbc7d39a253f11.

apeschar:
Let's continue the discussion on this in #30.

PR #37, opened 2021-03-09
Run formatters in file directory, not project directory

apeschar:
This relates to issue #36.

I believe that it's generally not good to change the working directory of the formatter to the project directory, because underlying directories will often contain conflicting or additional formatter configurations.


raxod502:
It was easier to simply revert the change adding this feature, so I did that in https://github.com/raxod502/apheleia/commit/87115f5cda2a6c0653fc0ac69bbbc7d39a253f11.

PR #40, opened 2021-06-18
Update apheleia-formatters type spec to avoid customize-variable error

mavant:
Declare all of the permitted special symbols used in apheleia-formatters commands.

Without this, 'customize-variable never allows settting the value of apheleia-formatters,
instead reporting an error about a mismatch between the default value and the type.


raxod502 (approved):
Thank you, this looks great! Merging.

Issue #44, opened 2021-08-08
Integrating `rustfmt`


adimit:
# Background

I wanted to use Apheleia for Rust code, too, as `rustic-mode`'s own formatter is rather buggy and nowhere near as good as Apheleia's handling.

However, `rustfmt` behaves in a way that confuses Apheleia. The existing PR #24 is not going to integrate it successfully. There are several problems:

- `rustfmt` defaults to checking all dependent files, not just the file it is passed. This means that its output with the `--emit stdout` option contains multiple files. To output only one file, there's an unstable `--skip-children` flag, which requires us to also allow such flags.
- (in my eyes at least) `--skip-children` has a bug: it still outputs the file name in the first line, and two additional newlines.

# Questions to Apheleia

- Is there a provision to make changes to the output of formatting commands? I'd like to drop the first two lines without shell magic.

Of course, I think that the underlying issue should be patched out of `rust-analyzer`. I'll see if I can do that.

# Workaround

In the meantime, I've created a workaround. First, you need to register `rustfmt` with Apheleia:

```emacs-lisp
    (add-to-list 'apheleia-formatters
                 '(rustfmt  . ("rustfmt-helper" file)))
    (add-to-list 'apheleia-mode-alist
                 '(rustic-mode . rustfmt)
``` 

I've made a shell script called `rustfmt-helper` to work around the above issues:

```bash
#!/bin/bash

rustfmt --emit stdout --unstable-features --skip-children $1 | tail -n +3
exit ${PIPESTATUS[0]}
```

Put this on your `$PATH` and `chmod +x` it.

adimit:
And of course I just found a solution without a shell script right after posting this: adding the `--quiet`-flag to `rustfmt` should get it to shut up about which file it is outputting. This should obviate the need for a test script. I'll check if that works and make a PR here, as the currently merged `rustfmt` specification is broken.

adimit:
Everything checks out AFAICS, I've created a PR.

raxod502:
Merged, sorry for the delay.

PR #45, opened 2021-08-08
Fix rustfmt dumping child modules' content into file

adimit:
See #44 for details.

rustfmt by default analyses all files dependent on the current file,
too. The output of `--emit stdout` contains all child modules. Apheleia
would happily dump that into the currently edited buffer (even though
the content came from another file.)

The unstable option --skip-children prevents that. It's as yet
untested! I've only made a couple of quick sanity checks.


npajkovsky:
This should be merged or fixed other way ASAP. apheleia is practically useless for rust.

raxod502 (approved):
:sparkles: Thanks for the fix and sorry it waited so long.

PR #50, opened 2021-10-05
Evaluate any non-string, non-special args in apheleia-formatters

joshforged:
file, input, and output are special cases in apheleia-formatters that
are specially expanded when interpreted by apheleia-format-buffer.
This commit makes it so any other symbols will be evaluated.

### DIFF
--- a/apheleia.el
+++ b/apheleia.el
@@ -448,7 +448,11 @@ modified from what is written to disk, then don't do anything."
                                   arg))
                               command)))
       (apheleia--make-process
-       :command command
+       :command (mapcar (lambda (arg)
+                          (if (stringp arg)
+                              arg
+                            (eval arg)))
+                        command)
        :stdin (unless input-fname
                 (current-buffer))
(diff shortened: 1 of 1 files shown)


lassik:
I advise against this. This is potentially unsafe if the values come from `.dir-locals.el`.

raxod502:
This is perfectly safe. The whole notion of safe file-local variables exists for exactly this reason, and Emacs will refuse to automatically apply a file-local setting for `apheleia-formatters` unless it is approved by the user. Besides, you could already `(setf (alist-get prettier apheleia-formatters) '("sudo" "something-malicious"))`; this doesn't make things any worse.

The implementation looks fine; could you add a changelog entry and update the docstring?

mohkale:
Is there anything I can do to help move this along? I'm running into a need for this with shfmt with regards to current indentation (tabs or space-width) and shell type (bash, sh, zsh, etc.)

PR #51, opened 2021-10-16
Support multiple formatters (#31)

### LINKED ISSUE #31: Allow running multiple formatters in sequence
First, thank you for this great package. 

Unless I'm mistaken, there is currently no way to configure `apheleia-mode` to run multiple formatters on save. When working with Python files, I often run `isort` followed by `black`. It would be nice to automate this workflow using this package. Maybe the `apheleia-mode-alist` could be changed to, for instance:
```emacs-lisp
((python-mode . (isort black))
 (js-mode . prettier))
```
and the package would run the list entries sequentially.

mohkale:
Closes #31

This commit makes it so apheleia can run multiple formatters one after
the other and use the resultant output to format the current buffer.
This works somewhat like a pipeline. The output of one formatter becomes
the input to the next formatter until all formatters have run and then
an RCS patch is built from the resultant output and applied to the
current buffer.

Note: For convenience we internally represent the users configuration as
a list of formatters even when it may only be one. For example if the
user has configured `(python-mode . black)` in apheleia-mode-alist then
internally we interpret that as a formatter list of `(black)` instead of
`black` as we did previously.


mohkale:
Some noteworthy changes:
+  Some function names I've renamed internally since they reference the singular but now should be pluralised. Example `apheleia--run-formatters` instead of `apheleia--run-formatter`.
+ `apheleia--run-formatters` is recursive now. The callback for the first formatter calls the function again with the rest of the formatters setting the stdin for the next formatter to the current stdout.
+ I've abstracted the conversion from a formatter command (list of symbols and strings) into a separate function `apheleia--format-command`.
+ I've made `apheleia--run-formatters` delete the temporary files in the callback. This should avoid leaving them alive unintentionally however since the callback is only run when the process exits succesfully the files will probably still be left-over if the command fails. Might be worth adding some sort of `:ensure` parameter to `apheleia--make-process` that's in charge of removing the files after the formatter exits.
+

Before we merge there's some stuff I'd like to confirm and clarify. I've added several comments to relevent sections which I'll link to here for convenience:
+ [apheleia--format-command](https://github.com/mohkale/apheleia/blob/662f1df2d28986be35c86bffe512f815abb5ca9a/apheleia.el#L404): The 'file predicate.
+ [apheleia--run-formatters](https://github.com/mohkale/apheleia/blob/662f1df2d28986be35c86bffe512f815abb5ca9a/apheleia.el#L476): Does lexical binding ensure the current-buffer is bound?

README.md
```diff
@@ -83,10 +83,30 @@ variables:
...
   regexps to names of formatters to use in those modes and files. See
   the docstring for more information.
+    * You can use this variable to configure multiple formatters for
+      the same buffer by setting the `cdr` of an entry to a list of
+      formatters to run instead of a single formatter. For example you
+      may want to run `isort` and `black` one after the other.
+
+      ```elisp
+      (push '(isort . ("isort" "--stdout" "-"))
+            apheleia-formatters)
```
raxod502, line 96: Would it make sense to use `(setf (alist-get 'isort apheleia-mode-alist) ...)` here as well?

apheleia.el
```diff
@@ -389,18 +389,29 @@ as its sole argument."
...
+  ;; file so it doesn't apply the formats from multiple formatters being
+  ;; run. For example if formatter 1 produces some output and formatter
+  ;; 2 modifies file in place then the changes from formatter 1 may not
+  ;; apply at all.
   (cl-block nil
     (let ((input-fname nil)
           (output-fname nil)
           (npx nil))
+      ;; TODO: Support arbitrary package managers, not just npx.
+      ;; TODO: Should we do this only when npx is at the car of cmd?
```
raxod502, line 414: We're not actually calling into NPM here, this symbol is just used for setting `default-directory` from Elisp. The behavior could certainly be generalized, though, and probably should be.

The current behavior is that the same thing happens whether `npx` is at the beginning, or if it appears anywhere in the list. Unless there's a compelling reason otherwise, I'd suggest keeping that behavior for compatibility.

apheleia.el
```diff
@@ -472,8 +518,26 @@ modified from what is written to disk, then don't do anything."
     (terraform . ("terraform" "fmt" "-")))
   "Alist of code formatting commands.
 The keys may be any symbols you want, and the values are
-commands, lists of strings and symbols, in the format of
-`apheleia-format-buffer' (which see)."
+commands, lists of strings and symbols.
+
+In Lisp code, commands is similar to what you pass to
```
raxod502, line 523:
```suggestion
In Lisp code, the format of commands is similar to what you pass to
```

apheleia.el
```diff
@@ -472,8 +518,26 @@ modified from what is written to disk, then don't do anything."
...
+want its contents to be passed on stdin.) If you instead use the
+symbol `input' as one of the elements of commands, then the
+contents of the current buffer are written to a temporary file
+and its name is substituted for `input'. Also, if you use the
+symbol `output' as one of the elements of commands, then it is
+substituted with the name of a temporary file. In that case, it
+is expected that the command writes to that file, and the file is
+then read into an Emacs buffer. Finally, if you use the symbol
+`npx' as one of the elements of commands, then the first string
+element of commands is resolved inside node_modules/.bin if such a
```
raxod502, line 539:
```suggestion
element of the command list is resolved inside node_modules/.bin if such a
```

apheleia.el
```diff
@@ -519,24 +583,39 @@ symbols (matched against `major-mode' with `derived-mode-p') or
...
 
 Be careful when writing regexps to include \"\\'\" and to escape
 \"\\.\" in order to properly match a file extension. For example,
 to match \".jsx\" files you might use \"\\.jsx\\'\"."
   :type '(alist
-          :key-type symbol
-          :value-type symbol))
+          :key-type
+          (choice (symbol :tag "Major mode")
+                  (string :tag "Buffer name regexp"))
```
raxod502, line 597: Thanks for catching this oversight.

apheleia.el
```diff
@@ -519,24 +583,39 @@ symbols (matched against `major-mode' with `derived-mode-p') or
...
+                  (repeat
+                   (symbol :tag "Formatter")))))
 
 (defvar-local apheleia-formatter nil
   "Name of formatter to use in current buffer, a symbol or nil.
 If non-nil, then `apheleia-formatters' should have a matching
 entry. This overrides `apheleia-mode-alist'.")
 
-(defun apheleia--get-formatter-command (&optional interactive)
+(defmacro apheleia--ensure-list (arg)
```
raxod502, line 608: Does this need to be a macro? It seems like it would function identically as, excuse the pun, a function.

apheleia.el
```diff
@@ -519,24 +583,39 @@ symbols (matched against `major-mode' with `derived-mode-p') or
...
+  "Ensure ARG is a list of length at least 1.
+When ARG is not a list its turned into a list."
+  `(when-let ((val ,arg))
+     (if (listp val)
+         val
+       (list val))))
+
+(defun apheleia--get-formatter-commands (&optional interactive)
   "Return the formatter command to use for the current buffer.
```
raxod502, line 617:
```suggestion
  "Return the formatter commands to use for the current buffer.
```

apheleia.el
```diff
@@ -545,28 +624,32 @@ If INTERACTIVE is non-nil, then prompt the user for which
...
                            (cl-dolist (entry apheleia-mode-alist)
                              (when (or (and (symbolp (car entry))
                                             (derived-mode-p (car entry)))
                                        (and (stringp (car entry))
                                             buffer-file-name
                                             (string-match-p
                                              (car entry) buffer-file-name)))
-                               (cl-return (cdr entry))))))
+                               (cl-return
+                                (apheleia--ensure-list (cdr entry)))))))
```
raxod502, line 638: It feels like we could hoist the call to `apheleia--ensure-list` outside of the enclosing `or`, and thereby avoid duplicating the call.

apheleia.el
```diff
@@ -583,48 +666,29 @@ even if a formatter is configured."
...
-element of COMMAND is resolved inside node_modules/.bin if such a
-directory exists anywhere above the current `default-directory'.
-
-In any case, after the formatter finishes running, the diff
-utility is invoked to determine what changes it made. That diff
-is then used to apply the formatter's changes to the current
-buffer without moving point or changing the scroll position in
-any window displaying the buffer. If the buffer has been modified
-since the formatter started running, however, the operation is
-aborted.
```
raxod502, line 619: It would be nice if in this docstring we had a pointer saying that the format is explained in the docstring of `apheleia-formatters`.

CHANGELOG.md
```diff
@@ -4,6 +4,9 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog].
 
 ## Unreleased
+### Enhancements
+* Support multiple formatters ([#31]).
```
raxod502, line 8: Can we include a brief explanation of this feature? (You can now use a list in `apheleia-mode-alist`, and all the formatters listed will be run in sequence.)

apheleia.el
```diff
@@ -389,18 +389,29 @@ as its sole argument."
...
-(defun apheleia--run-formatter (command callback)
-  "Run a code formatter on the current buffer.
-The formatter is specified by COMMAND, a list of strings or
-symbols (see `apheleia-format-buffer'). Invoke CALLBACK with one
-argument, a buffer containing the output of the formatter.
-
-If COMMAND uses the symbol `file' and the current buffer is
-modified from what is written to disk, then don't do anything."
+(defun apheleia--format-command (command &optional stdin)
+  "Format COMMAND into a shell-ccmd and list of file paths.
```
raxod502, line 393: Is `ccmd` a typo or just an abbreviation I'm not familiar with?

apheleia.el
```diff
@@ -389,18 +389,29 @@ as its sole argument."
...
+Should return a list with the car being the optional input file-name
```
raxod502, line 394:
```suggestion
Return a list with the car being the optional input file-name
```

apheleia.el
```diff
@@ -389,18 +389,29 @@ as its sole argument."
...
+the formatters standard input.
+
+If COMMAND uses the symbol `file' and the current buffer is modified
+from what is written to disk, then return nil meaning meaning no
+cmd is to be run."
+  ;; WARN: The condition with regards to `file' only checks the original
+  ;; file so it doesn't apply the formats from multiple formatters being
+  ;; run. For example if formatter 1 produces some output and formatter
+  ;; 2 modifies file in place then the changes from formatter 1 may not
+  ;; apply at all.
```
raxod502, line 408: Ideally, we can handle all supported features for formatter chaining. But failing that, I think it would be good enough to report an error or warning when we hit an unsupported case.

apheleia.el
```diff
@@ -447,15 +459,49 @@ modified from what is written to disk, then don't do anything."
...
+STDIN is a buffer containing the standard input for the first
+formatter in COMMANDS. This should not be supplied by the caller
+and instead is supplied by this command when invoked recursively.
+The stdout of the previous formatter becomes the stdin of the
+next formatter."
+  ;; WARN: Is there a chance that current-buffer will change when invoked
+  ;; recursively? Say when first called current-buffer is the input file
+  ;; buffer, when called again is it still the original file buffer?
+  ;;
+  ;; To summarise: Does emacs persists current-buffer with lexical scoping?
```
raxod502, line 481: Yeah, so in principle the current buffer in the callback should be the stdout buffer associated with the process, regardless of what the current buffer was when the process was spawned.

However, it's probably best to explicitly save and track the current buffer in a lexically scoped variable to avoid any surprises. `current-buffer` is dynamically scoped, regardless of whether lexical binding is enabled or not. (Well, actually `current-buffer` is a function, not a variable, but if you were to analogize it to a variable, it would behave more like a dynamically scoped one than a lexically scoped one.)

raxod502 (commented):
Thanks! Made some comments and hopefully answered your questions, let me know if I can help with anything else.

README.md
mohkale: I suppose so, I wasn't aware setf actually pushed new entries onto the list as well. I thought if alist-get returns nil it would just fail but I just tested it out and it works like it should. In the meantime what're your thoughts on adding isort to `apheleia-formatters`?

apheleia.el
mohkale: Typo. It was originally just shell-command but the documentation linter complained it wasn't uppercase and I had to condense it (clearly didn't do a good job) I'll fix that now.

apheleia.el
mohkale: I thought the docstring near the start of the function did that. Or would you like it to be a little more descriptive?

```
COMMANDS is a list of the values from `apheleia-formatters'. 
```

raxod502:
Pushed my last commit to the wrong upstream. I got in the right branch this time.

> I'm slightly confused by the distinction between file and filepath

Good point, it does look like they're equivalent in the current code, which is a bug. However, there _should_ be a distinction, we shouldn't be sending stdin for `file`. The use of `file` should indicate that the formatter MUST read from the actual file, whereas `filepath` just means it needs the filename, but can read from either stdin or a temporary file, as appropriate.

mohkale:
Okay, kind of ugly but I'm passing stdin to `apheleia--format-command` and then shadowing the value of stdin in `apheleia--run-formatters` with the value returned be `apheleia--format-command`. Since the file or filepath section in `apheleia--format-command` may set stdin to nil, but stdin is needed by input, I've moved it to the end of the function. That should be everything for now.

PR #59, opened 2021-11-20
Support non-file buffers (#52)

### LINKED ISSUE #52: Better handling for non-file buffers
Atm apheleia has `filepath` and `file` command specifiers that're replaced with the value of the current buffers file. Quite a lot of the time I'm working with virtual files (such as by creating an org src-block for some c code) and apheliea will refuse to format that buffer (when called interactively through `apheleia-format-buffer`) because its been modified and it doesn't have a file-name. There should be a way to format buffers as well as files.

mohkale:
Closes #52

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ The format is based on [Keep a Changelog].
   be run in sequence.
 * Support evaluating items in `apheleia-formatters` to make formatter
   commands more dynamic ([#50], [#55]).
+* Allow apheleia to format buffers without an underlying file ([#52])
 
 ### Formatters
 * [ClangFormat](https://clang.llvm.org/docs/ClangFormat.html) for
@@ -46,6 +47,7 @@ The format is based on [Keep a Changelog].
 [#48]: https://github.com/raxod502/apheleia/pull/48
 [#49]: https://github.com/raxod502/apheleia/pull/49
 [#50]: https://github.com/raxod502/apheleia/pull/50
+[#52]: https://github.com/raxod502/apheleia/issues/52
 [#54]: https://github.com/raxod502/apheleia/pull/54
 [#55]: https://github.com/raxod502/apheleia/issues/55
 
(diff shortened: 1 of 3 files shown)


CHANGELOG.md
```diff
@@ -10,6 +10,7 @@ The format is based on [Keep a Changelog].
   be run in sequence.
 * Support evaluating items in `apheleia-formatters` to make formatter
   commands more dynamic ([#50], [#55]).
+* Allow apheleia to format buffers without an underlying file ([#52])
```
raxod502, line 13:
```suggestion
* Allow apheleia to format buffers without an underlying file ([#52]).
```

README.md
```diff
@@ -107,6 +107,18 @@ variables:
       only when it is bound. Observe that one of these evaluations
       returns a list of flags whereas the other returns a single
       string. These are substituted into the command as you'd expect.
+    * You can also use apheleia to format buffers that have no underlying
```
raxod502, line 110:
```suggestion
    * You can also use Apheleia to format buffers that have no underlying
```

apheleia.el
```diff
@@ -394,6 +397,14 @@ as its sole argument."
                       ;; changes, and 2 if error.
                       (memq status '(0 1)))))))
 
+(defun apheleia--safe-buffer-name ()
+  "Current `buffer-name' without special file-system characters."
+  ;; See [[https://stackoverflow.com/questions/1976007/what-characters-are-forbidden-in-windows-and-linux-directory-names][here]] for a list of supported characters on all systems.
```
raxod502, line 402:
```suggestion
  ;; See https://stackoverflow.com/q/1976007 for a list of supported 
  ;; characters on all systems.
```

apheleia.el
```diff
@@ -394,6 +397,14 @@ as its sole argument."
                       ;; changes, and 2 if error.
                       (memq status '(0 1)))))))
 
+(defun apheleia--safe-buffer-name ()
+  "Current `buffer-name' without special file-system characters."
```
raxod502, line 401:
```suggestion
  "Return `buffer-name' without special file-system characters."
```

raxod502 (approved):
:sparkles: Seems legit. Thank you!

PR #63, opened 2021-12-11
[#62] Support functions as formatters

### LINKED ISSUE #62: Functional formatters (for use with lsp-mode/eglot)
I've noticed apheleia atm doesn't support functional formatters, such as `pp-buffer`. 

I'd like to add support for using language servers (eglot, or lsp-mode) for formatting a buffer on save through apheleia so I'd like apheleia to support functional formatters. 

I'm interested in what your thoughts on this would be. Implementation wise it might be more convenient to clone the current buffer, or copy it and then run the formatter on the cloned buffer and continue with the changes we were already doing. This would also allow chaining with the existing formatter system. So we could say format with lsp-mode, then run python-isort, etc.

Interface wise I was thinking we just allow the cdr of an entry in `apheleia-formatters` to be a symbol, and if it is then its a function to call. Luckily atm their all lists, even gofmt.

mohkale:
Closes #62.

Lets you use a lisp function as a formatter. This gives apheleia a lot
more flexibility in regards to what constitutes a formatter. For example
you can now plug an external language server or another tool as a
formatter for use with apheleia.

Here's a very basic example of using indent-line-function with apheleia
after merging this commit. Note: this doesn't take into account any
special local variables in the original buffer such as lisp-body-indent.
It's really just for demonstration purposes and as a proof of concept.

```lisp
(defun apheleia-indent-region+ (orig scratch callback)
  (with-current-buffer scratch
    (setq-local indent-line-function
                (buffer-local-value 'indent-line-function orig))
    (indent-region (point-min)
                   (point-max))
    (funcall callback scratch)))

(push '(indent-region . apheleia-indent-region+) apheleia-formatters)
(push '(elisp-mode . indent-region) apheleia-mode-alist)
(push '(lisp-interaction-mode . indent-region) apheleia-mode-alist)
```

### DIFF
--- a/apheleia.el
+++ b/apheleia.el
@@ -543,12 +543,7 @@ description of COMMAND, BUFFER, CALLBACK and STDIN."
              (erase-buffer)
              (insert-file-contents-literally output-fname))
 
-           (if (cdr commands)
-               ;; Forward current stdout to remaining formatters, passing along
-               ;; the current callback and using the current formatters output
-               ;; as stdin.
-               (apheleia--run-formatters (cdr commands) buffer callback stdout)
-             (funcall callback stdout)))
+           (funcall callback stdout))
          :ensure
          (lambda ()
            (ignore-errors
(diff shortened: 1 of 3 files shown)


mohkale:
@raxod502 

Just one question. How does apheleia handle multiple calls to format in a short period? Say I've got 5 python files open and modified and then I save them all. Since they all have the same formatter configuration, wouldn't apheleia overwrite the stdout of some of the formatters with other formatters? How do you ensure consistency and that the formatters don't step over each other?

apheleia.el
```diff
@@ -517,20 +508,64 @@ next formatter."
...
+         (ignore-errors
            (when output-fname
              (delete-file output-fname))))))))
 
+(defun apheleia--run-formatter-function (func buffer callback stdin)
+  "Run a formatter using a Lisp function FUNC.
+See `apheleia--run-formatters' for a description of BUFFER, CALLBACK
+and STDIN."
+  ;; Will be an ugly name if you use a lambda for FUNC, instead of a symbol.
+  (let ((scratch (get-buffer-create (format " *apheleia-%s-scratch*" func))))
```
raxod502, line 526: Might this have a race condition similar to what you fixed in your other PR, in case multiple formatter functions are running in parallel? We might want to implement the same fix from there (and then might as well make the buffer name prettier, e.g. by using `lambda` when the function is not a symbol).

raxod502 (commented):
Implementation-wise, this seems great, the only thing it needs is the merge conflict resolved, and maybe a sentence in the README mentioning that this feature exists, and pointing at the existing docstring.

apheleia.el
mohkale: Ah, yep, nice spot.

mohkale:
@raxod502 

Okay, that should've fixed the previous issues. I've also modified the functional formatter specification. It takes an extra callback to call when an error is raised and it doesn't pass the formatted buffer to callback anymore. It's now assumed the scratch buffer is what will  be formatted.

raxod502 (approved):
:sparkles:

PR #74, opened 2022-01-08
Define apheleia-formatter as a safe-local-variable

terlar:
This adds support for variables from .dir-locals for users that are
using the setting enable-local-variables :safe or it won't ask for those
who have it set to t.

Should I add this note to the changelog as well?

Something like:

> * You can now specify  `apheleia-formatter` inside `.dir-locals` when
  using `enable-local-variables` `:safe` or it won't ask you if you have
  it set to `t`.


raxod502:
Looks good to me, changelog update sounds good, I added it as suggested. Will merge in a moment.

PR #76, opened 2022-01-11
Support formatting remote files with Tramp

### LINKED ISSUE #33: TRAMP support
As noted in the README, TRAMP is currently unsupported. I am considering adding support since I use TRAMP extensively. I'm interested in your thoughts on this before I potentially embark on creating a PR. :slightly_smiling_face:

mohkale:
Adds support for formatting remote files.

The new `apheleia-remote-algorithm` option configures this. The default
behaviour is consistent with what we had before, with `apheleia`
aborting a formatting when dealing with a remote file/buffer.
Users can customise apheleia to run the formatter on the remote machine,
in which case any temporary files or other checks such as `npx` will be
fully handled on the remote machine.
Users can also make apheleia run the formatter on the local machine.
This works exactly like one would expect, except if the formatter
requires access to the physical file (meaning it uses 'file in
`apheleia-formatters`) because the file isn't available on the local
machine.

This PR also fixes a bug in apheleia where `input-fname` was assigned in
local let scope, instead of being returned by `apheleia--format-command`,
meaning any formatters using a 'input file weren't cleaned up after
formatting.

Closes #33.


mohkale:
@raxod502 

I've tested this by creating ssh-ing into my own machine within emacs and using tramp. I've found both `cancel` and `local` options work (which is what I was really after) but `remote` doesn't. It runs all the commands and then gets to the point where it runs the diff command but the diff just hangs forever. I'm not sure whether this is an emacs issue or a diff issue (maybe it's doing something funky with the stdin and tramp doesn't get along with that). It's worth trying to figure out why this is happening, although it's not a big concern for me atm.

mohkale:
I also found and fixed a bug where if we created a file for the `diff` generation, we didn't then delete it when we were done.

raxod502:
Thank you! I made a couple of stylistic changes; do those look okay to you? If so then I can go ahead and merge this.

raxod502:
Also, looks like linting was failing on Emacs 25 because `make-temp-file` didn't previously have three arguments. Since Emacs 26 has been available since May 2018, over 3.5 years ago, I think it is okay to just drop support at this point.

mohkale:
> If so then I can go ahead and merge this.

Probably should wait until we've addressed why running formatter processes on the remote hangs sometimes. That's still an issue. I can run the process on the local machine for remote buffers, but if I try to run it on the remote machine it never manages to finish. The other day it hanged at the formatter stage, and not the diff one, so I'm not sure what's really causing it.

We could also disable the running on remote functionality, and merge in running apheleia on remote buffers with local processes (that's honestly fine for my use-case), but I'd like to fix this if we can.

raxod502:
This is actually the bug that is more concerning to me. When formatting a remote buffer on save, I get this:

image

And then if saying yes, then the following backtrace:

<details>
<summary>Backtrace</summary>

```
Debugger entered--Lisp error: (remote-file-error "Forbidden reentrant call of Tramp")
  signal(remote-file-error ("Forbidden reentrant call of Tramp"))
  tramp-error(#<process *tramp/sshx venli*> remote-file-error "Forbidden reentrant call of Tramp")
  tramp-send-string((tramp-file-name "sshx" nil nil "venli" nil "/home/raxod502/control/messenger-mirror/" nil) "test -e /home/raxod502/control/messenger-mirror/ 2...")
  tramp-send-command((tramp-file-name "sshx" nil nil "venli" nil "/home/raxod502/control/messenger-mirror/" nil) "test -e /home/raxod502/control/messenger-mirror/ 2...")
  tramp-send-command-and-check((tramp-file-name "sshx" nil nil "venli" nil "/home/raxod502/control/messenger-mirror/" nil) "test -e /home/raxod502/control/messenger-mirror/")
  tramp-sh-handle-file-exists-p("/sshx:venli:/home/raxod502/control/messenger-mirro...")
  tramp-sh-file-name-handler(file-exists-p "/sshx:venli:/home/raxod502/control/messenger-mirro...")
  apply(tramp-sh-file-name-handler file-exists-p "/sshx:venli:/home/raxod502/control/messenger-mirro...")
  tramp-file-name-handler(file-exists-p "/sshx:venli:/home/raxod502/control/messenger-mirro...")
  file-exists-p("/sshx:venli:/home/raxod502/control/messenger-mirro...")
  basic-save-buffer(nil)
  #<subr save-buffer>()
  funcall(#<subr save-buffer>)
  (let ((apheleia-mode (and apheleia-mode (member arg '(nil 1))))) (funcall func))
  radian--save-buffer-reformat-maybe(#<subr save-buffer>)
  apply(radian--save-buffer-reformat-maybe #<subr save-buffer> nil)
  save-buffer()
  apheleia--save-buffer-silently()
  #<subr F616e6f6e796d6f75732d6c616d626461_anonymous_lambda_36>()
  #f(compiled-function (patch-buffer) #<bytecode 0x97826ec6c9b66ed>)(#<buffer  *apheleia-diff-stdout*>)
  #f(compiled-function (proc event) #<bytecode -0x1c9d67586d30ec5f>)(#<process apheleia-diff> "exited abnormally with code 1\n")
  apply(#f(compiled-function (proc event) #<bytecode -0x1c9d67586d30ec5f>) (#<process apheleia-diff> "exited abnormally with code 1\n"))
  #f(advice-wrapper :after #f(compiled-function (proc event) #<bytecode -0x1c9d67586d30ec5f>) #f(compiled-function (proc msg) #<bytecode 0x100a91790850564e>))(#<process apheleia-diff> "exited abnormally with code 1\n")
  accept-process-output(#<process *tramp/sshx venli*> nil nil t)
  tramp-accept-process-output(#<process *tramp/sshx venli*>)
  tramp-wait-for-regexp(#<process *tramp/sshx venli*> nil "\\(^\\|\0\\)[^#$\n]*///d144d484e021d666559eff64993fe5b5...")
  tramp-wait-for-output(#<process *tramp/sshx venli*>)
  tramp-send-command((tramp-file-name "sshx" nil nil "venli" nil "/tmp/tramp.upjYVS" nil) "test -d /tmp/tramp.upjYVS 2>/dev/null; echo tramp_...")
  tramp-send-command-and-check((tramp-file-name "sshx" nil nil "venli" nil "/tmp/tramp.upjYVS" nil) "test -d /tmp/tramp.upjYVS")
  tramp-run-test("-d" "/sshx:venli:/tmp/tramp.upjYVS")
  tramp-sh-handle-file-directory-p("/sshx:venli:/tmp/tramp.upjYVS")
  tramp-sh-file-name-handler(file-directory-p "/sshx:venli:/tmp/tramp.upjYVS")
  apply(tramp-sh-file-name-handler file-directory-p "/sshx:venli:/tmp/tramp.upjYVS")
  tramp-file-name-handler(file-directory-p "/sshx:venli:/tmp/tramp.upjYVS")
  #<subr delete-file>("/sshx:venli:/tmp/tramp.upjYVS")
  apply(#<subr delete-file> "/sshx:venli:/tmp/tramp.upjYVS")
  delete-file("/sshx:venli:/tmp/tramp.upjYVS")
  #f(compiled-function (proc msg) #<bytecode 0x10274c4ee250564e>)(#<process apheleia-black> "finished\n")
  apply(#f(compiled-function (proc msg) #<bytecode 0x10274c4ee250564e>) (#<process apheleia-black> "finished\n"))
  #f(advice-wrapper :after #f(compiled-function (proc event) #<bytecode 0x14bbc819598defed>) #f(compiled-function (proc msg) #<bytecode 0x10274c4ee250564e>))(#<process apheleia-black> "finished\n")
```

</details>

This occurs only when the formatting operation needs to change the buffer.

raxod502:
It looks like what's happening there is that since we are running the formatter asynchronously, but all TRAMP commands share a common connection process, we try to `delete-file` which should be synchronous, but this is instead submitted over TRAMP, and while we are waiting for a response, we find that the formatter has finished, and try to run the process sentinel, which involves executing more commands over TRAMP, which is reentrant usage.

I'd like to take a look at a couple other issues before investigating this further, but thought I would drop my findings here.

mohkale:
@raxod502 

Yeah, damn, I'm encountering this issue as well. From this [issue](https://www.mail-archive.com/tramp-devel@gnu.org/msg05623.html) it looks like there isn't much library writers can do to work around this atm. On the apheleia side we might be able to pool pending writes into a queue and only write one at a time, but no solution seems very practical from my POV. Might be worth closing this issue

raxod502:
@PythonNut I know you've done a lot of work with TRAMP and were the original person to request remote file formatting support; any thoughts on if there's a good workaround for this issue?

mohkale:
@raxod502 

What do you think about making apheleia blocking if run remotely? The main issue at this stage is the user could modify the file whilst its being formatted or a myriad of other tramp issues stepping over each other. If apheleia locked emacs while it was formatting it should fix most of those issues, and that's better than just refusing to run altogether or throwing errors while saving.

raxod502:
I mean, if we previously didn't have any support for remote formatting, and we add support in a degraded mode, then it seems like an improvement to me. I think anything involving tramp always involves some synchronous waiting, unfortunately, just due to the way they've designed things. So it's kind of par for the course.

Be warned though, I've tried on a number of occasions over the years to synchronously wait for `make-process` to return, and have never managed to implement it in a way that didn't have noticeable race conditions. See e.g. https://github.com/raxod502/straight.el/issues/334 although that's not the only place I tried it if I recall correctly.

You're probably best off adding a higher-level interface that will delegate calls either to `make-process` or `call-process` depending on whether synchronous or asynchronous operation is desired.

mohkale:
@raxod502 

Okay, I got started on it but this is kind of a nightmare :scream:. 

The interface for `call-process` is so bizarre but I can workaround most of the quirks except for the fact there's no way to use the `file-name-handler` with a synchronous process execution that takes a stdin buffer. There's
+ `call-process` which doesn't take STDIN and doesn't use file-handler.
+ `process-file` which doesn't take STDIN, but does use a file-handler
+  `call-process-on-region` which takes stdin and doesn't use a file-handler.
+ but there's no `process-file-on-region` that can do both.

That means if there's a STDIN buffer and we're running the formatter remotely then there's no way to execute it and pass standard input without first saving stdin on the remote and then doing some funky sort of subshell redirection.

mohkale:
What I have now should work, but it's failing with black because TRAMP seems to be including some junk at the end of the copied file :/.

apheleia.el
```diff
@@ -289,7 +319,109 @@ at once.")
...
+                          (shell-quote-argument
+                           (apheleia--strip-remote remote-stdin)))))
+                   (unwind-protect
+                       (progn
+                         (with-current-buffer stdin
+                           (apheleia--write-region-silently nil nil
+                                                            remote-stdin))
+
+                         (process-file
+                          shell nil (nth 2 args) nil "-c" shell-command))
```
mohkale, line 406: Copying happens here, black is saying `error: cannot format -: '\x1b'` and the remote file ends with a `^[[0m` for some reason/

raxod502:
Thanks! I should be able to take a look at this next weekend and see if I can help resolve the outstanding issues.

apheleia.el
raxod502: Weird -- one thing I'll note is `\x1b[0m` is the ANSI control sequence for "reset color" -- maybe TRAMP adds that for some reason because it's trying to reuse the same shell session and doesn't want the color to be messed up by a poorly behaved program?

raxod502:
Interesting, what you have now actually does seem to work for me as is! It takes forever to do all the round-tripping, but I am able to format a Python file over TRAMP using the `sshx` connection method with `(setq apheleia-remote-algorithm 'remote)` and Black 19.3b0 installed as `/usr/bin/black` on the remote.

Here's a sample Python file that formats successfully using this method: https://github.com/raxod502/radian/blob/58ba58bd827e719c0eeb3d3c996d59cf4d00acd5/scripts/print_env.py

Do you perhaps have a different buffer coding system? Maybe you are using the `ssh` connection method in TRAMP instead of `sshx`, and your shell configuration or terminal prompt is adding some ANSI escape sequences that TRAMP is failing to parse correctly? I always found the `ssh` method to be a trainwreck and use `sshx` instead because it doesn't depend on any customization on the shell side.

mohkale:
Hmm... I'm using sshx (ssh has pretty never worked for me). Maybe it's because I'm configured to use both black and isort with apheleia. I'll try with just black a little later and see whether it's a chained formatter issue.

raxod502:
Hmm, that's odd, I still can't reproduce. I set

```elisp
(setq apheleia-remote-algorithm 'remote)
(setf (alist-get 'python-mode apheleia-mode-alist) '(isort black))
```

and verified that `(apheleia--get-formatters)` in the remote file returns `(isort black)`. Then I took the same sample file from before, and rearranged the imports to be out of the correct order. When I saved, the imports were reordered correctly, and also the long lines were wrapped by Black, and I didn't see any errors or extraneous output added to the file. This is with black 19.3b0 and isort 4.3.4.

Maybe you have a different version of Black? I know Black likes to use escape sequences to colorize stuff and do emojis and stuff, so maybe somehow that's getting triggered in your setup and not mine?

mohkale:
I have `black, 22.1.0 (compiled: yes)` and `isort VERSION 5.9.2`, and I'm still encountering this issue :disappointed:. Downgrading my isort version to yours seems to have fixed the issue so I'm guessing its an isort bug. Looking into the help there is a `--color` option (but no option to disable it, kinda weird since I don't see a need for it anyways). Either way this PR seems to be working which is good. I'd like to disable the kinda meaningless `Renaming /sshx:kisara:/tmp/tramp.mjX5Um to /tmp/apheleiapRYi1m...done` messages that're triggered by `process-lines` but I'm not sure how to.

raxod502:
> I'd like to disable the kinda meaningless `Renaming /sshx:kisara:/tmp/tramp.mjX5Um to /tmp/apheleiapRYi1m...done` messages that're triggered by `process-lines` but I'm not sure how to.

You might be able to get rid of this by customizing the value of `tramp-verbose`, perhaps automatically in `let`. If that doesn't do the trick, I would use `cl-letf` to temporarily rebind `(symbol-function #'message)` so that it checks and ignores messages that match the relevant pattern. I do this often enough in Radian that [I have a special macro for it](https://github.com/raxod502/radian/blob/91239bbd6fad24a3271d638429835b4f5bef569b/emacs/radian.el#L131-L151).

mohkale:
> You might be able to get rid of this by customizing the value of tramp-verbose, perhaps automatically in let.

I tried that before commenting it but sadly it didn't work. My guess is it's not tramp outputting the message, it's rename-file or whatever other command is calling tramp under the hood.

>  I would use cl-letf to temporarily rebind (symbol-function #'message) so that it checks and ignores messages that match the relevant pattern.

Good suggestion, I'll try that out :-).

mohkale:
Yep, I managed to suppress the annoying messages. 

It took a while, and is slow as hell, but we can finally format remote buffers with apheleia. I'm hoping docker-tramp is much more performant than sshx for this (since that's what I'm planning to use it with). Let me know if there's anything else you'd like to get worked on with this PR, otherwise I think it's ready to merge :-).

mohkale:
Update: I just noticed `apheleia-remote-algorithm = 'local` still has the reentrant call of tramp issue. Guess we'll have to make that blocking as well.

mohkale:
Okay, what I have no should work but the code is really tangly so it might need a good refactoring to make it easier to understand.

apheleia.el
```diff
@@ -913,8 +1133,11 @@ even if a formatter is configured."
 
 (defun apheleia--disallowed-p ()
   "Return an error message if Apheleia cannot be run, else nil."
-  (when (and buffer-file-name (file-remote-p buffer-file-name))
-    "Apheleia does not support remote files"))
+  (when (and buffer-file-name
+             (file-remote-p (or buffer-file-name
+                                default-directory))
+             (eq apheleia-remote-algorithm 'cancel))
+    "Apheleia refused to run formatter due to `apheleia-remote-algorithm'."))
```
raxod502, line 1140:
```suggestion
    "Apheleia refused to run formatter due to `apheleia-remote-algorithm'"))
```

apheleia.el
```diff
@@ -690,6 +901,8 @@ formatter being run, for diagnostic purposes."
                  (unwind-protect
                      (funcall callback scratch)
                    (kill-buffer scratch)))
+               ;; A metadata field
+               `((remote . ,remote))
```
raxod502, line 905: Is this changing the expected interface of Elisp-based formatters? If so we would want to update the docstring of `apheleia-formatters` and announce a breaking change.

If we are making a breaking change anyway, it might be a good time to make the interface extensible, by saying that the function must receive its parameters using keyword arguments, and ignore any unknown keys.

raxod502 (commented):
Aside from my one comment below, this looks great, thank you!

antifuchs:
I'm extremely excited about this change - it might turn apheleia into the single autoformatting emacs mode that can deal with remote files (:

So, I tried this out and there's a bit of an annoying issue that I haven't debugged fully yet: Accessing a linux box from a macOS-based emacs, it tries to write files to a macOS-specific temporary directory (which isn't set up on the linux side):

```
Tramp: Encoding local file ‘/var/folders/r3/z_pbvnkx60j5kbfb8bw6kjz40000gn/T/tramp.cBNzGC’ using ‘base64-encode-region’...done
Tramp: Decoding remote file ‘/scpx:ferdl.hack:/var/folders/r3/z_pbvnkx60j5kbfb8bw6kjz40000gn/T/apheleia-stdinSrkw2j’ using ‘base64 -d -i >%s’...failed
```

I suspect there's a problem here...

raxod502:
> wonder if there's a way to make emacs generate a reasonable temp file on the remote machine

Could invoke `mktemp` as a subprocess, would be slower but more reliable.

antifuchs:
> Could invoke mktemp as a subprocess, would be slower but more reliable.

So, funny story, I looked at `make-nearby-temp-file` and it seems that this does pretty much exactly what the [`apheleia--make-temp-file`](https://github.com/raxod502/apheleia/pull/76/files#diff-66ff3951c9eb3afbe61be90c0933b5f16906b815f1a4874383c1a37160a89217R633) does, without the necessary remote tempdir customization.

mohkale:
> it might turn apheleia into the single autoformatting emacs mode that can deal with remote files

Deal with very slowly, but yes that is the hope 😂.

> I looked at `make-nearby-temp-file`

Good spot, I didn't notice this existed. I'll update the PR to use that over an apheleia specific wrapper.

apheleia.el
mohkale: @raxod502 

It looks like functional formatters are still unreleased (in the CHANGELOG), so is a breaking change still necessary?

mohkale:
@raxod502 

Okay. I've tried refactoring (as best as I can) but its mostly identical since the high level flow between the functions is the same. I did rename some of the process oriented functions to make their behaviour a little clearer (but it might be just as confusing :sweat:).

I've adapted `make-nearby-temp-file` into `apheleia--make-temp-file` and removed the custom path option. I've also updated the interface of functional formatters to take key-value arguments (adding a new one for whether the formatter should be run asynchronously or not as well).

I've tested all these changes using isort+black both locally and on a remote machine with sshx using both `apheleia-remote-algorithm` as local and remote. One final thing @raxod502 you should be aware of as the maintainer is that the wrapper I've written for call-process and make-process can both be run on a remote and locally. Currently the functionality/branch where we run a formatter asynchronously on a remote machine is never triggered (because of aforementioned race conditions) but should tramp ever provide a workaround for this we should relatively easily be able to incorporate them.

I'd suggest maybe keeping this PR open for a little while to test (since we've had multiple issues prop up in actual usage) before merging. But otherwise, fingers crossed, I think this is finally ready :weary:.

antifuchs:
I pulled this in, removed my `apheleia-remote-temporary-file-directory` customization and ran the isort&black formatters on a remote python file - and it seems to work just fine!

I'll work with this for few more days (working on nix, prettier-formatted typescript, yaml and rustfmt-formatted rust); if something should look weird, I'll let you know here.

mohkale:
Hmmm... I'm encountering this error while running apheleia locally (not through tramp).

```lisp
Debugger entered--Lisp error: (error "process apheleia-diff no longer connected to pipe; closed it")
  process-send-string(#<process apheleia-diff> #("\"\"\"\nConnect to the database and get infor..." 0 2 (face (tree-sitter-hl-face:doc tree-sitter-hl-face:string) fontified nil) 2 3 (face (tree-sitter-hl-face:doc tree-sitter-hl-face:string) fontified nil syntax-table (15)) 3 152 (face (tree-sitter-hl-face:doc tree-sitter-hl-face:string) fontified nil) 152 153 (face (tree-sitter-hl-face:doc tree-sitter-hl-face:string) fontified nil syntax-table (15)) 153 155 (face (tree-sitter-hl-face:doc tree-sitter-hl-face:string) fontified nil) 155 157 (fontified nil) 157 163 (face tree-sitter-hl-face:keyword fontified nil) 163 176 (fontified nil) 176 182 (face tree-sitter-hl-face:keyword fontified nil) 182 192 (fontified nil) 192 198 (face tree-sitter-hl-face:keyword fontified nil) 198 204 (fontified nil) 204 210 (face tree-sitter-hl-face:keyword fontified nil) 210 221 (fontified nil) 221 227 (face tree-sitter-hl-face:keyword fontified nil) 227 236 (fontified nil) ...))
  apheleia--make-process(:name "apheleia-diff" :stdin #<buffer group_config.py> :stdout #<buffer  *apheleia-diff-stdout*> :stderr #<buffer  *apheleia-diff-stderr*> :command ("diff" "--rcs" "--strip-trailing-cr" "--" "/home/mkaleem2/prog/repos/..." "/tmp/apheleiaiJNsm7") :remote nil :noquery t :callback #f(compiled-function (proc-exit-status) #<bytecode -0xfed67df59ad9fc5>))
  apheleia--execute-formatter-process(:command ("diff" "--rcs" "--strip-trailing-cr" "--" "/home/mkaleem2/prog/repos/..." "/tmp/apheleiaiJNsm7") :stdin #<buffer group_config.py> :callback #f(compiled-function (patch-buffer) #<bytecode -0x6d3b376c77c055b>) :remote nil :ensure #f(compiled-function () #<bytecode -0x1fa365b9d834b280>) :exit-status #f(compiled-function (status) #<bytecode -0x1d95bbb5ac5fb746>))
  apheleia--create-rcs-patch(#<buffer group_config.py> #<buffer  *apheleia-black-stdout*> nil #f(compiled-function (patch-buffer) #<bytecode -0x6d3b376c77c055b>))
  #f(compiled-function (formatted-buffer) #<bytecode 0x13d4b009a176819d>)(#<buffer  *apheleia-black-stdout*>)
  #f(compiled-function (stdout) #<bytecode -0x1c46680c0673027b>)(#<buffer  *apheleia-black-stdout*>)
  #f(compiled-function (stdout) #<bytecode -0xa45069ec37881d3>)(#<buffer  *apheleia-black-stdout*>)
  #f(compiled-function (proc-exit-status) #<bytecode -0x846ca31eb9b4a11>)(0)
  #f(compiled-function (proc event) #<bytecode 0x1c56358a7501e31d>)(#<process apheleia-black> "finished\n")
```

I'll try loading apheleia directly (without byte compiling) to get a more helpful error message.

mohkale:
I've pushed a fix for the above issue and tested locally. Can't test on a remote ATM so I'd appreciate someone else doing that.

raxod502:
Works for me on a remote machine with both the `local` and `remote` algorithms! Thanks @mohkale!!

gsingh93:
The README still says TRAMP isn't supported, but I tested TRAMP support out and it seems to work. Can the README be updated and some documentation about `apheleia-remote-algorithm` be added?

PR #90, opened 2022-04-26
Add shfmt

elken:
Add support for shfmt. _Hopefully_ there's a neater way to do the installer, I'm not a gopher.

### DIFF
--- a/apheleia.el
+++ b/apheleia.el
@@ -922,6 +922,7 @@ being run, for diagnostic purposes."
     (ocamlformat . ("ocamlformat" "-" "--name" filepath
                     "--enable-outside-detected-project"))
     (prettier . (npx "prettier" "--stdin-filepath" filepath))
+    (shfmt . ("shfmt"))
     (rustfmt . ("rustfmt" "--quiet" "--emit" "stdout"))
     (terraform . ("terraform" "fmt" "-")))
   "Alist of code formatting commands.
@@ -1045,6 +1046,7 @@ function: %s" command)))
     (rustic-mode . rustfmt)
     (rust-mode . rustfmt)
     (sass-mode . prettier)
+    (sh-mode . shfmt)
     (terraform-mode . terraform)
     (TeX-latex-mode . latexindent)
     (TeX-mode . latexindent)
--- a/test/formatters/install-common.bash
+++ b/test/formatters/install-common.bash
@@ -4,7 +4,7 @@ set -euxo pipefail
 
 export DEBIAN_FRONTEND=noninteractive
 apt-get update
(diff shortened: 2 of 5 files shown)


apheleia.el
```diff
@@ -922,6 +922,7 @@ being run, for diagnostic purposes."
     (ocamlformat . ("ocamlformat" "-" "--name" filepath
                     "--enable-outside-detected-project"))
     (prettier . (npx "prettier" "--stdin-filepath" filepath))
+    (shfmt . ("shfmt" "-ci"))
```
raxod502, line 925: Should we use the default settings by default, and let people customize it if they want to add additional options? Or do you think most people would want `-ci` so we should include it by default?

test/formatters/apheleia-ft.el
```diff
@@ -260,7 +260,7 @@ environment variable, defaulting to all formatters."
                  in-file)
                (list stdout-buffer stderr-file)
                nil
-               (cdr command)))
+               (cdr `(,command))))
```
raxod502, line 263: Isn't the `cdr` of a list with one element always going to be nil?

raxod502 (commented):
Thanks! This looks good, I pushed a commit that simplifies the installation (in my experience, Go tooling is atrocious and it's almost always better to just download a separate binary release if one is available), and I think all we need is a changelog entry.

apheleia.el
elken: Yeah you're right here, we shouldn't make decisions by default.

test/formatters/apheleia-ft.el
elken: Hm yes you're right, this is a leftover change from a previous impl (I tried to use `sh-shell` and `indent-tabs-mode`) so I can remove it now I'm using a flat list

elken:
> I pushed a commit that simplifies the installation

The issue with that is it doesn't take into consideration other CPU archs (which is why I opted to go for the build in the first place)

raxod502:
Upstream offers binary downloads in all of the architectures that your switch statement covers; does the latest commit work for you?

elken:
`latest_release` doesn't run for me, and I'm also not on linux :p So I need the `uname` snippet

raxod502:
Those scripts are intended to be run inside Docker as part of the Apheleia build system (`make fmt-build FORMATTERS=shfmt`). Are you running them manually?

elken:
I didn't think so, but I'll check again later

(Otherwise this is probably fine to merge)

raxod502:
Installation succeeds and tests for shfmt are passing in CI at https://github.com/radian-software/apheleia/runs/6397390532, so I'll merge.

PR #94, opened 2022-05-09
Use ocamlformat for caml-mode too

elken:
Also use `ocamlformat` for [caml-mode](https://github.com/ocaml/caml-mode)


raxod502 (approved):
:sparkles: LGTM with added changelog entry.

PR #98, opened 2022-05-14
Add nixfmt

michzappa:


raxod502:
Thanks, I fixed a Markdown issue in the changelog (and also one from a previous pull request), the code all looks good.

Issue #99, opened 2022-06-01
Renaming default branch from 'master' to 'main'


raxod502:
Following https://github.com/github/renaming, I've renamed the default branch on Apheleia from `master` to `main`. You can update your local copy following [these instructions](https://docs.github.com/en/repositories/configuring-branches-and-merges-in-your-repository/managing-branches-in-your-repository/renaming-a-branch):

```
git branch -m master main
git fetch origin
git branch -u origin/main main
git remote set-head origin -a
```

Let me know if this change causes any disruption and I can see what I can do to mitigate!

raxod502:
Pushed https://github.com/radian-software/apheleia/commit/b25256f1b014ee02887c7882a789f95dc87d061b to fix CI errors cc/ @elken and I've updated your PR branches, sorry for the inconvenience.

elken:
Think this is probably ok to close now?

raxod502:
Yep, sorry for delay.

PR #101, opened 2022-06-03
Add bean-format

elken:
Might also require some kind of deindent formatter (may be that we
introduce some generic formatters like that to chain other against?)
but otherwise should work fine


elken:
Test will likely fail due to #99

test/formatters/installers/bean-format.bash
```diff
@@ -0,0 +1,2 @@
+apt install -y python3-pip
```
raxod502, line 1:
```suggestion
apt-get install -y python3-pip
```

test/formatters/installers/bean-format.bash
```diff
@@ -0,0 +1,2 @@
...
+pip install beancount
```
raxod502, line 2:
```suggestion
pip3 install beancount
```

raxod502:
Interesting, what's the indentation problem? Does the formatter not work well when applied to an already-formatted file?

elken:
It's just a very simple formatter that doesn't deindent, even the tests just use python `textwrap` to do that.

raxod502:
What do you mean by `deindent`? When I hear that word I think of reducing the indentation level of a block of code, but your reference to the `textwrap` module makes me think you're talking about hard-wrapping long lines of text into paragraphs.

elken:
I do indeed mean reducing indentation, perhaps the tests are misusing that module then.

All this does is align columns of numbers, if the indentation is wonky it doesn't adjust it.

raxod502:
I think it's fine to leave as is. If formatting this file format requires additional functionality then it should be added into the formatter, not hacked on top in an Emacs package!

PR #102, opened 2022-06-03
Add emacs-lisp formatting

elken:
Created as a draft because it doesn't _quite_ work; I've tried a couple of approaches but they either don't update correctly or require a second write.

I'm _sure_ I'm implementing it incorrectly, so this is more an RFC than an actual changeset :)


dakra:
There is an example to re-indent elisp in https://github.com/radian-software/apheleia/pull/63

elken:
> There is an example to re-indent elisp in #63

That's where I start, doesn't quite work :)

apheleia.el
```diff
@@ -927,6 +927,7 @@ being run, for diagnostic purposes."
     (gofmt . ("gofmt"))
     (google-java-format . ("google-java-format" "-"))
     (isort . ("isort" "-"))
+    (el-indent-buffer . apheleia-indent-emacs-lisp-buffer)
```
raxod502, line 930: To make it immediately obvious what this is for
```suggestion
    (emacs-lisp-indent . apheleia-indent-emacs-lisp-buffer)
```

apheleia.el
```diff
@@ -998,6 +999,12 @@ rather than using this system."
              (const :tag "Name of temporary file used for output" output)))
            (function :tag "Formatter function"))))
 
+(cl-defun apheleia-indent-emacs-lisp-buffer (&key buffer formatter callback remote stdin &allow-other-keys)
```
raxod502, line 1002: Should have docstring

apheleia.el
```diff
@@ -998,6 +999,12 @@ rather than using this system."
...
+  (with-current-buffer buffer
+    (save-excursion
+      (funcall indent-region-function (point-min) (point-max))
+      (save-buffer 1))))
```
raxod502, line 1006: I don't think you want to invoke `save-buffer` here.

raxod502:
What are the issues you're facing with your current implementation? Maybe I can help advise.

elken:
> What are the issues you're facing with your current implementation? Maybe I can help advise.

You're right that I don't want to use save-buffer, but if I don't then the buffer is left changed. If I try and use the method in #63 the callback throws a wrong arguments error

elken:
Pushed changes, this now works in emacs but the test fails for the _slightest_ diff

```diff
diff --git a/tmp/apheleia-ft-file-EDn7yr.expected b/tmp/apheleia-ft-file-Uvnxkx.actual
index 7aeecb4..c0d102c 100644
--- a/tmp/apheleia-ft-file-EDn7yr.expected
+++ b/tmp/apheleia-ft-file-Uvnxkx.actual
@@ -1,4 +1,4 @@
 (if (and (< 3 5)
-         (= 1 1))
+        (= 1 1))
     (message "true")
   (message "false"))
```

raxod502:
> there's an issue with tabs vs spaces

Yeah, that's just because `indent-tabs-mode` foolishly defaults to non-nil. We can disable it in file-local variables, no need to pull in anything from Doom. (That function is just for fixing up plist indentation, which is a definite improvement but just not the issue at hand here.) I've pushed a commit fixing the issue.

apheleia.el
```diff
@@ -984,19 +985,34 @@ repository is automatically prepended to $PATH (variable
...
-             (string :tag "Argument")
-             (const :tag "Look for command in node_modules/.bin" npx)
-             (const :tag "Name of file being formatted" filepath)
-             (const :tag "Name of real file used for input" file)
-             (const :tag "Name of temporary file used for input" input)
-             (const :tag "Name of temporary file used for output" output)))
-           (function :tag "Formatter function"))))
+   :type '(alist
+           :key-type symbol
+            :value-type
```
raxod502, line 990: Hmm, what happened to this indentation?

elken:
Related to the comment, this makes the indentation _very_ wacky

image

Thoughts?

elken:
Technically, this should work for any lisp. Should I add tests for common lisp and maybe scheme too?

test/formatters/apheleia-ft.el
```diff
@@ -218,65 +218,82 @@ environment variable, defaulting to all formatters."
...
+             (append `(,(expand-file-name
+                         "scripts/formatters"
+                         (file-name-directory
+                          (file-truename
+                           ;; Borrowed with love from Magit
+                           (let ((load-suffixes '(".el")))
+                             (locate-library "apheleia"))))))
+                     exec-path)))
         (setq stdout-buffer (get-buffer-create
                              (format "*apheleia-ft-stdout-%S" formatter)))
```
raxod502, line 230: Why do we do this unconditionally now?

test/formatters/apheleia-ft.el
```diff
@@ -218,65 +218,82 @@ environment variable, defaulting to all formatters."
...
+                (funcall command
+                         :buffer (current-buffer)
+                         :scratch (current-buffer)
+                         :formatter formatter
+                         :callback (lambda ()))
+                (copy-to-buffer stdout-buffer (point-min) (point-max))))
+          (progn
+
+            (with-current-buffer stdout-buffer
+              (erase-buffer))
```
raxod502, line 247: Why do we move this higher up now?

test/formatters/apheleia-ft.el
elken: I didn't change this :P

elken:
OK scratch that, I was tired when I did that.

Apart from the merge conflict I'm about to resume, are you happy with this?

elken:
@raxod502 what's left to do here? the lint failure doesn't align with how the function lints elisp.

Think we should adjust the lint step to utilize what we define here so it dogfoods itself. If desired, I can add that to this PR

raxod502:
I feel bad for letting this sit so long, let me see if I can fix up any remaining CI failures and merge this in.

PR #104, opened 2022-06-13
Add lua-fmt

kisaragi-hiu:
Add [lua-fmt].

(Why the formatter is called `lua-fmt` despite the executable being `luafmt` (no dash): [lua-fmt] and [luafmt](https://github.com/CurtisFenner/luafmt) are two separate projects. The one added here is what [format-all](https://github.com/lassik/emacs-format-all-the-code/) supports by default.)

[lua-fmt]: https://github.com/trixnz/lua-fmt/


elken:
Looks like lua-fmt hasn't been updated since 2018, I wouldn't recommend making this the default. Instead, look into stylua or LuaFormatter (I use the former)

kisaragi-hiu:
stylua seems to be recommended more. I'll look into submitting that instead.

kisaragi-hiu:
Closing in favor of https://github.com/radian-software/apheleia/pull/105.

PR #105, opened 2022-06-18
Add stylua

kisaragi-hiu:
Add stylua, [which is a more up-to-date formatter than lua-fmt](https://github.com/radian-software/apheleia/pull/104#issuecomment-1158573448).

stylua uses tabs by default while spaces seem more popular for Lua. However, passing `--indent-type Spaces --indent-width 2` would override the user's `.stylua.toml`, so I've kept its default.


elken:
I think leaving it as default is fine if it overrides the config, most projects should have one anyway. And if not, it's quite simple to override the behaviour

raxod502 (approved):
:sparkles: Looks great, thanks!

PR #107, opened 2022-06-20
[#103] Specify --parser in Prettier

### LINKED ISSUE #103: prettier formatter error:  find no parser
I want to format a file name: .yamllint, it is definitely a yaml file. However, the prettier format 
returns `find no parser` error even if I have set the major mode to `yaml-mode`.

It seems prettier select parser according to file extension like .vue, .json or whatever
```bash
# prettier --help
--parser <flow|babel|babel-flow|babel-ts|typescript|acorn|espree|meriyah|css|less|scss|json|json5|json-stringify|graphql|markdown|mdx|vue|yaml|glimmer|html|angular|lwc|toml>
```
```elisp
(prettier . (npx "prettier" "--stdin-filepath" filepath))
```
I notice that prettier format doesn't include the `--parser=` argument, could we map
(MAJOR_MODE . PARSER) by force or select parser according to file extension first and then the major-mode ?

raxod502:
Closes #103


lassik:
See also https://github.com/prettier/prettier/issues/13052

elken:
WDYT about the linked PR @raxod502 ?

This should also be expanded to handle GraphQL and Markdown

raxod502:
The linked PR has an interesting idea. Since we already have to explicitly support each new major mode, I am not sure if it will actually improve the user experience to adopt that approach versus what I have here. Let me add Markdown and GraphQL as you suggested (subject to https://github.com/radian-software/apheleia/pull/81), then merge this.

raxod502:
Filed https://github.com/ruby-syntax-tree/syntax_tree/issues/144 for an upstream issue I discovered while trying to get CI to pass.

elken:
Oh how annoying!

raxod502:
The folks upstream most generously fixed that issue, but apparently the same problem occurs with https://github.com/ruby-syntax-tree/syntax_tree-rbs, which is also a dependency, so I'll submit a pull request for that.

raxod502:
I filed https://github.com/ruby-syntax-tree/syntax_tree-rbs/pull/34, but my Ruby inexperience is showing, unfortunately.

raxod502:
Okay, I fixed up https://github.com/ruby-syntax-tree/syntax_tree-rbs/pull/34 and confirmed that it fixes the CI here, so after that PR is merged, I'll go ahead and rebuild this and it should be good to go.

If it takes a long time to merge the upstream PR, then I can adjust the CI here to pull from my branch for now.

raxod502:
Sorry this took so long. But it's done now!

PR #110, opened 2022-07-05
Allow formatter using 'filepath on a modified buffer

### LINKED ISSUE #109: Can’t run Prettier on modified buffer via M-x apheleia-format-buffer
I want to be control of whether/when my code gets formatted, so I want to directly use `apheleia-format-buffer` rather than the automatic save hook. However, if I modify a JavaScript buffer without saving it, then run `M-x apheleia-format-buffer`, nothing happens—the buffer is not formatted.

This is because `apheleia--format-command` returns `nil` here:

https://github.com/radian-software/apheleia/blob/12804a50206c708570104637472411288d6133f5/apheleia.el#L779-L789

The apparent assumption is that any command including `file` or `filepath` must be reading the file from disk, and therefore won’t work on a modified buffer. But that’s a false positive in Prettier’s case—the command is `(npx "prettier" "--stdin-filepath" filepath)`, and Prettier is only using `filepath` to determine which parser to use and locate configuration options. This seems to have been the intention behind the difference between `file` and `filepath`.

andersk:
Fixes #109.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,8 @@ The format is based on [Keep a Changelog].
 ## Unreleased
 ### Enhancements
 * shfmt uses 4 spaces instead of tabs by default.
+* Formatters using `'filepath` (OCamlFormat and Prettier) are no
+  longer prevented from running on a modified buffer.
 
 ### Formatters
 * [bean-format](https://github.com/beancount/beancount) for Beancount
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,7 +7,7 @@ The format is based on [Keep a Changelog].
 ### Enhancements
 * shfmt uses 4 spaces instead of tabs by default.
 * Formatters using `'filepath` (OCamlFormat and Prettier) are no
-  longer prevented from running on a modified buffer.
+  longer prevented from running on a modified buffer ([#109], [#110]).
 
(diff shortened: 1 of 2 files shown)


mohkale:
I don't think this is a good idea. The restriction from formatting when the buffer is modified is because it assumes the formatter is formatting the file, not the buffer, and then when we try to apply the changes atop the buffer because it doesn't match what was there before formatting there could be some discrepencies. That said I'm also not a big fan of the current state of this. It might be better to add an option to auto-save the buffer before running the formatter or add a new field (aside from filepath) which indicates to apheleia "yes, this uses the filepath, but it still formats through stdin".

andersk:
`filepath` *is* the option that indicates to apheleia “yes, this uses the filepath, but still formats through stdin”. If you want to avoid indicating that, you use `file` instead of `filepath`. See this comment:

https://github.com/radian-software/apheleia/blob/12804a50206c708570104637472411288d6133f5/apheleia.el#L959-L961

elken:
I can see this argument and I'm kinda in favour of it.

The thought of making the formatting run more transparently during typing is an interesting one.

raxod502:
You're right. There is no need to block formatting on a modified buffer for `filepath`, only for `file`.

raxod502 (approved):
:sparkles:

PR #114, opened 2022-08-07
Added fantomas for fsharp

mattiasdrp:
I couldn't find a way to make it work with stdin so it takes the file and outputs the result to stdout

I did all the lines asked in the contributor guide and it worked for me :-)


elken:
Looks like there are 2 flags for stdin

    --fsi <string>        Read F# source from stdin as F# signatures.
    --stdin               Read F# source from standard input.

Did these not work? If not this should probably be reported upstream

raxod502:
This thread is being closed automatically by [Tidier](https://github.com/radian-software/tidier) because it is labeled with "waiting on response" and has not seen any activity for 90 days. But don't worry—if you have any information that might advance the discussion, leave a comment and I will be happy to reopen the thread :)

PR #116, opened 2022-08-17
Only process when there is output to return

elken:
scalafmt only returns stdout when there are errors, this could also help
in other cases where a formatter returns incorrectly.

Ideally, this would also be improved by basic error checking heuristics
to work out if we need to update the buffer.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,10 @@ The format is based on [Keep a Changelog].
 * Formatters using `'filepath` (OCamlFormat and Prettier) are no
   longer prevented from running on a modified buffer ([#109], [#110]).
 
+### Bugs fixed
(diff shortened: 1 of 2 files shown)


raxod502 (approved):
:sparkles: Thanks. I updated the changelog accordingly. In my opinion, if a formatter does this then it is a bug that should be fixed upstream, but it is undeniable that the change you've suggested here is a simple way to improve the user experience in Apheleia irrespective of such bugs.

elken:
> :sparkles: Thanks. I updated the changelog accordingly. In my opinion, if a formatter does this then it is a bug that should be fixed upstream, but it is undeniable that the change you've suggested here is a simple way to improve the user experience in Apheleia irrespective of such bugs.

In the case of scalafmt it seems to be by design rather than a bug 😅 

But yeah I do agree

PR #119, opened 2022-08-21
Use :connection-type pipe when creating local processes

### LINKED ISSUE #118: [Solved] Standard Prettier config fails, how to debug it? Logs are insufficient. Can't figure out what causes the exit code 1.
The default `prettier` config doesn't work for me. It always returns with an error about the exit code:

```
Wed Aug 17 17:39:39 2022 :: /umetra/test/my-app/src/
$ /umetra/test/my-app/node_modules/.bin/prettier --stdin-filepath /umetra/test/my-app/src/hooks.ts

(no output on stderr)

Command failed with exit code 1.
```

I tried to change it by providing a parser and removing the guesswork, still doesn't work.

```
Wed Aug 17 17:34:05 2022 :: /umetra/test/my-app/src/
$ /umetra/test/my-app/node_modules/.bin/prettier --parser\=typescript

(no output on stderr)

Command failed with exit code 1.
```

From the command line, it works fine with any of the options above if I provide the file from STDIN like this:
`cat src/hooks.ts | npx prettier --parser=typescript`

It also works fine when I provide a `parser` and `filepath`. But in this case, it adds some gibberish coming from the STDOUT:
```ts
hooks.ts[2K[1Gimport cookie from 'cookie';
import { v4 as uuid } from '@lukeed/uuid';
import type { Handle } from '@sveltejs/kit';
```

I'm lost about debugging what causes Aphelia to see an exit code of 1 while it works on the command line. There's a feeling that STDIN is not there for some reason, it's the only case how I managed to reproduce an exit code. What I can do from here to make it work again?

leotaku:
This is helpful because some formatters, rather than having an option
to enable reading from stdin, simply check whether their input comes
from a pipe or terminal. I don't really think this is a wise thing for any code formatter/application to do, but unfortunately it seems somewhat widespread, so we should support it.

In particular, this fixes prettier, so should close #118.

I hope this suits you code and documentation-style wise, wasn't really how to best handle adding a new but non-changing `make-process` argument with the keyword-argument forwarding the existing code does.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,9 @@ The format is based on [Keep a Changelog].
 * shfmt uses 4 spaces instead of tabs by default.
 * Formatters using `'filepath` (OCamlFormat and Prettier) are no
   longer prevented from running on a modified buffer ([#109], [#110]).
+* Buffer content is now always passed to formatters using a pipe. This
+  fixes issues with formatters that behave differently when receiving
+  input on stdin versus being run on a tty.
 
 ### Formatters
 * [bean-format](https://github.com/beancount/beancount) for Beancount
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,7 +10,7 @@ The format is based on [Keep a Changelog].
   longer prevented from running on a modified buffer ([#109], [#110]).
 * Buffer content is now always passed to formatters using a pipe. This
   fixes issues with formatters that behave differently when receiving
-  input on stdin versus being run on a tty.
+  input on stdin versus being run on a tty ([#119]).
 
 ### Formatters
 * [bean-format](https://github.com/beancount/beancount) for Beancount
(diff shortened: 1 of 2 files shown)


leotaku:
The long argument lists seem to be causing lint errors. Please instruct me on how you want this fixed, as it seems like this mostly depends on your personal taste.

raxod502 (approved):
:sparkles: Thanks! I went ahead and broke the long lines into two which fixes the linter. Will merge after CI.

artemkovalyov:
It's a fantastic solution, @leotaku, I missed that option of `make-process` somehow.

lassik:
All formatters should be able to do stdin->stdout via pipes. If a formatter requires a tty, that's a bug in the formatter.

leotaku:
@lassik It's actually the other way around. The `prettier` command tests if its input is connected to a pipe, and if this is the case, waits for data from stdin. If it thinks its input is connected to a terminal, it requires a filename as an argument to read input from.

While, as I've said, I don't think this is a good decision, it is in fact deliberate and not a bug as evidenced by prettier/prettier#4559.

lassik:
I agree the new behavior is brittle, but still it seems using pipes will always work.

PR #126, opened 2022-09-08
Guard against missing formatters

### LINKED ISSUE #122: Check if formatter exists before calling it
With the default configuration, if `global-apheleia-mode` is enabled you get an error message like this when saving a file when the formatter is not installed:
```
Failed to run shfmt: Searching for program: No such file or directory, shfmt
```

I think it would be better to check if the formatter exists first and then emit a warning if it doesn't. I think it's nice that the default value for `apheleia-mode-alist` has various formatters already configured, so if I ever do want to start using a formatter with a language I can just install it and things should work. But I often set `debug-on-error` to debug issues in various packages, so I keep getting this error popuping up whenever I save a file where I don't have the formatter installed (and I don't need or want it installed, at least at the moment).

elken:
Fixes #122

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -25,6 +25,10 @@ The format is based on [Keep a Changelog].
   running. Instead, it is disabled in narrowed buffers. Support for
   narrowed buffers may be added in future but it has never been
   correctly supported in the past (see [#43]). More at [#124], [#127].
+* Currently, when a formatter invoked via a process isn't installed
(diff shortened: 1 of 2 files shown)


raxod502 (approved):
:sparkles: OK, fair enough.

raxod502:
Mind adding changelog and merging?

elken:
> Mind adding changelog and merging?

Sure!

PR #128, opened 2022-09-08
Assume filename correctly

### LINKED ISSUE #108: clang-format fails to determine main header
With the following `_clang-format`

```
SortIncludes: true
IncludeBlocks: Regroup
```
`clang-format -i sort_includes.cpp` formats the following `sort_includes.cpp`:
```
#include "other_include.hpp"
#include <vector>
#include "sort_includes.hpp"
```
into 
```
#include "sort_includes.hpp"

#include "other_include.hpp"

#include <vector>
```
However, running `aphelia-format-buffer` produces file with
```
#include "other_include.hpp"
#include "sort_includes.hpp"

#include <vector>
```
It seems that when apheleia runs `clang-format`, `clang-format` cannot determine the main header (see, https://clang.llvm.org/docs/ClangFormatStyleOptions.html IncludeCategories). I'm guessing this occurs because apheleia send the file to `clang-format` via standard input and for that reason `clang-format` does not know the filename. 

I was able to workaround the issue if I modified the `clang-format` entry in `apheleia-formatters` to `'("clang-format" file)`. `clang-format` also seems to have a command line option `--assume-filename=<string>`, which could also solve the issue.

elken:
Fixes #108


elken:
Test fails but functionally works, apheleia-ft needs to handle evaluating the formatter args

raxod502:
How does this change look to you?

raxod502:
I don't love switching on major modes inside the formatter command, but we don't have a better abstraction for doing this at the moment, so I think it is okay.

elken:
> I don't love switching on major modes inside the formatter command, but we don't have a better abstraction for doing this at the moment, so I think it is okay.

It defaults to the buffer name and only falls back for a temp buffer or something

elken:
> How does this change look to you?

LGTM! I'll push a changelog entry now

raxod502 (approved):
:sparkles:

PR #129, opened 2022-09-14
docs: add note about wiki

elken:
Refs: https://github.com/radian-software/apheleia/issues/115#issuecomment-1246138394


raxod502:
Thank you!

PR #132, opened 2022-10-04
Use inplace formatter output

leonidborisenko:
When formatting command is defined with symbol `inplace`, it doesn't work.

I'm using [casey/just](https://github.com/casey/just) (v1.5.0).

It doesn't support formatting of `stdin`, but it supports in-place formatting with command `just --unstable --fmt --justfile /tmp/justfile`. So I've added a list to to `apheleia-formatters` custom variable:

```elisp
(add-to-list 'apheleia-formatters
             '(just . ("just" "--unstable" "--fmt" "--justfile" inplace))
             "APPEND:non-nil")
```

Minimal `justfile` before formatting (there is one space before `echo`):
```make
rule:
 echo test
```

Expected formatting result (there are four spaces before `echo`):
```make
rule:
    echo test
```

Without this pull request there are no visible changes after <kbd>C-u</kbd> <kbd>M-x</kbd> `apheleia-format-buffer` <kbd>RET</kbd> `just` <kbd>RET</kbd>. When this pull request is applied, formatting proceed as expected.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -32,6 +32,8 @@ The format is based on [Keep a Changelog].
 * clang-format doesn't handle filenames correctly by default. Support
   for guessing the flags from first the file name then the major mode
   in the case of a temporary buffer was added in [#128]
+* Fix handling of formatters using `inplace` in some circumstances
+  ([#132]).
 
 ### Formatters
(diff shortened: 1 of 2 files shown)


raxod502:
Thanks!

PR #137, opened 2022-10-12
Add Crystal lang support.

zw963:
Thanks


zw963:
Hi, i have no idea for why CI not past, could you please help? thanks

raxod502:
This is what I get when I run locally:

```
apheleia@fc200f826997:/src$ make fmt-test FORMATTERS=crystal-tool-format
diff --git a/tmp/apheleia-ft-file-qepSxT.original b/tmp/apheleia-ft-file-WuYqMs.updated
index 5cf8a77..257743b 100644
--- a/tmp/apheleia-ft-file-qepSxT.original
+++ b/tmp/apheleia-ft-file-WuYqMs.updated
@@ -3,5 +3,5 @@ module SomeController
     def hello
       puts "hello"
     end
-      end           # <= this indentation is wrong.
+  end # <= this indentation is wrong.
 end
Formatter crystal-tool-format modified original file in place
make: *** [Makefile:97: fmt-test] Error 255
```

As the error message reports, the formatter is modifying the original file in place, which is not allowed. Can it read from stdin instead?

zw963:
> 

I guess there is no option for this.

```sh
 ╰─ $ cr tool format -h
Usage: crystal tool format [options] [file or directory]

Options:
    --check                          Checks that formatting code produces no changes
    -i <path>, --include <path>      Include path
    -e <path>, --exclude <path>      Exclude path (default: lib)
    -h, --help                       Show this message
    --no-color                       Disable colored output
    --show-backtrace                 Show backtrace on a bug (used only for debugging)
```

zw963:
All checks have passed!

raxod502 (approved):
:sparkles: Thanks!

PR #138, opened 2022-10-17
Introduce apheleia-inhibit-functions

### LINKED ISSUE #134: Conditionally disable apheleia-mode locally
When working in shared code base, auto-formatting may lead to diff
noise disasters when there isn't a commonly agreed code formatting
convention (or worse: people don't even care).

With `apheleia-global-mode` enabled, I think we should add an option
to conditionally disable auto-formatting that is ran at every save,
but still not get in the way when we want to call
`apheleia-format-buffer` explicitly.

While still using `apheleia-global-mode`, adding a "dummy" formatter
to `apheleia-formatters` (like `#'ignore`), and set
`apheleia-formatter` to it `.dir-locals-2.el` may disable
auto-formatting, but it prevents us from `M-x apheleia-format-buffer`.

daanturo:
Fixes #134

(I accidentally force-pushed #135 from a shallow repo)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,14 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog].
 
 ## Unreleased
+### Features
+* You can use `apheleia-inhibit` as a file-local variable to disable
+  Apheleia turning on automatically for a file or directory. You can
+  also use `apheleia-inhibit-functions` to configure custom logic to
+  prevent Apheleia from turning on automatically under certain
+  circumstances, without needing to adjust file-local variables. See
+  [#134] and [#138].
+
 ### Enhancements
 * shfmt uses 4 spaces instead of tabs by default.
 * Formatters using `'filepath` (OCamlFormat and Prettier) are no
@@ -71,6 +79,8 @@ The format is based on [Keep a Changelog].
 [#128]: https://github.com/radian-software/apheleia/pull/128
 [#132]: https://github.com/radian-software/apheleia/pull/132
 [#137]: https://github.com/radian-software/apheleia/pull/137
+[#134]: https://github.com/radian-software/apheleia/issues/134
+[#138]: https://github.com/radian-software/apheleia/pull/138
 
 ## 3.0 (released 2022-06-01)
 ### Breaking changes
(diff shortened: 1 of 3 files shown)


raxod502 (commented):
Thanks, this will be very helpful. I think we can improve the design a little bit, though. `define-globalized-minor-mode` already has a feature that is intended to accomplish conditional enabling of a minor mode:

```
Signature
(define-globalized-minor-mode GLOBAL-MODE MODE TURN-ON [KEY VALUE]... BODY...)

Documentation
Make a global mode GLOBAL-MODE corresponding to buffer-local minor MODE.

TURN-ON is a function that will be called with no args in every buffer
and that should try to turn MODE on if applicable for that buffer.
```

So, if we make the `TURN-ON` function check `apheleia--not-inhibited` before calling `apheleia-mode`, I think it will accomplish the same, but more simply.

I don't mind changing that in your PR directly but I wanted to ask you your thoughts first.

daanturo:
> So, if we make the `TURN-ON` function check `apheleia--not-inhibited` before calling `apheleia-mode`, I think it will accomplish the same, but more simply.

I think that's totally fine. For me, my condition is a bit slow to run so I originally want to defer checking until saving is triggered, but I can just modify it to add to `".dir-locals-2.el"` to evaluate just once anyway.

Cleaner code is superior, I'm in for `define-global-minor-mode`'s `TURN-ON`.

raxod502:
Ok, cool. Yeah, I figured that if the user desired caching then they could simply implement that in their disable-function.

raxod502:
How does this version look to you?

daanturo:
We may need to dump those variables into the autoloads file, else `apheleia-global-mode` will:
```
Error caused by user's config or system: ~/.config/doom/config.el, (void-variable apheleia-inhibit)
```

daanturo:
Ping @raxod502 , I added commit https://github.com/radian-software/apheleia/pull/138/commits/66f30de1795b95b2152e8d119ebbcd238f1daa84 .

raxod502:
Yeah, sorry for the delay, that seems fine to me. Would be equally okay to autoload the variable individually.

Looks like CI is failing, let me see if I can fix that.

raxod502:
Ah yes it's just checkdoc.

raxod502:
Thanks!

PR #142, opened 2022-11-10
feat: add deno fmt support

dalugm:
I'd like to pass `--ext` the argument with the file's extenstion. I tried to do this:

``` lisp
(deno-fmt . ("deno" "fmt" "-"
             "--ext" (or (cdr (assoc major-mode
                                     '((js-mode . "js")
                                       (js-jsx-mode . "jsx")
                                       (js-json-mode . "json")
                                       (markdown-mode . "md")
                                       (typescript-mode . "ts"))))
                         (or (file-name-extension (or buffer-file-name ""))
                             "ts"))
```

But it can't pass the CI test, any ideas?


raxod502:
I had a look locally, I assume this is what you are talking about?

```
apheleia@a7a9d42007c0:/src$ make compile fmt-test FORMATTERS=deno-fmt
[compile] apheleia.el
[format-test] success: formatter deno-fmt (file in.js)
error: Line 1, column 15: Expected ';', '}' or <eof>

  {"arrowParens":"always","
                ~
Formatter deno-fmt exited with status 1
make: *** [Makefile:97: fmt-test] Error 255
```

I think that is because the test framework does not set a major mode for the buffer. Nor necessarily should it, because in practice people could use multiple different major modes for a file type, depending on their personal preference, and what is supposed to be tested by the formatter tests is the operation of the formatter, not the major mode mapping.

I feel like it would be better to instead leave the major mode mapping to `apheleia-mode-alist`, whose job it is to map a major mode to a formatter configuration. In other words, I like what you have already.

Is `--ext` needed? Can we do the same thing as with `prettier`, where there are specialized formatters, but also one that uses the built-in formatter behavior?

dalugm:
>  this is what you are talking about?

Actually, I'd like to let the test framework keep the extension that the file has, then We don't need to specify all situations for deno.

> Is `--ext` needed? Can we do the same thing as with prettier, where there are specialized formatters, but also one that uses the built-in formatter behavior?

Yes, I noticed that if you are formatting a buffer without extension, deno will always assume it's formatting a ts file, which would lead to an error

图片

If deno is formatting a `a.json` file, then the `--ext` is not needed any more.

elken:
Yeah similar to clang-format where we have to do something similar, it sucks at guessing what the content you're passing actually is

dalugm:
I think write like this can solve most cases for deno, and avoid writing specialized formatters.

``` lisp
(deno-fmt . ("deno" "fmt" "-"
             "--ext" (or (file-name-extension (buffer-name))
                         "ts")
```

But it can't pass the CI, because CI compared with `xxx.original` and `xxx.updated`.

Using `in.json` and `out.json`, the problem should be solved, but I'm afraid it would lead to a huge change for the test framework.

dalugm:
> have something arguably wrong but working now

Well, I don't think current solution is wrong, I just thought the current solution is a bit heavy. In fact, either solution is good for me, I just want to write it as simple.

> you can do if you want to move it along faster

Only when I have enough time... And it will take a long time I guess since I'm not a programmer, I can only write some simple code based on the existing codes to fix some minor bugs.

PR #144, opened 2022-11-13
Add swift-format.

drshapeless:
Add swift-format.


elken:
Please resolve the CI issues and follow the [contributing guide](https://github.com/radian-software/apheleia#contributing)

drshapeless:
I could not find a easy way other than compiling from source to install swift-format in ubuntu.
Would that work if I clone the source and compile it in the bash script?

elken:
I'm no domain expert but it looks like it has a hard dependency on macos.

@raxod502 if this is the case, I assume we want to reject this?

drshapeless:
No, I am using it on Arch Linux. It doesn't require anything about macOS. But just the debian repo does not have this package available.

elken:
Sure that it's not [this](https://github.com/nicklockwood/SwiftFormat) instead? Can't see any way to build Apple's version without XCode.

raxod502:
If it's a formatter that's _only_ usable on macOS... yeah, that seems like the kind of thing that shouldn't be listed by default, because there's no way to make sure that future contributions to Apheleia don't break the functionality. If the only way to install it on Ubuntu is to compile from source, then yes, it's fine to compile and install from source in the installation script.

elken:
> If it's a formatter that's _only_ usable on macOS... yeah, that seems like the kind of thing that shouldn't be listed by default, because there's no way to make sure that future contributions to Apheleia don't break the functionality. If the only way to install it on Ubuntu is to compile from source, then yes, it's fine to compile and install from source in the installation script.

There seems to be two; Apple's own which seems to be MacOS only and the one I linked which isn't. Currently waiting on OP to let us know which they mean

raxod502:
This thread is being closed automatically by [Tidier](https://github.com/radian-software/tidier) because it is labeled with "waiting on response" and has not seen any activity for 90 days. But don't worry—if you have any information that might advance the discussion, leave a comment and I will be happy to reopen the thread :)

PR #145, opened 2022-11-22
apheleia-mode-alist: Add new tree-sitter based major modes

leungbk:
-------


raxod502 (commented):
Cool, makes sense, mind updating the changelog?

PR #146, opened 2022-12-02
Add prettier-markdown as formatter for markdown-mode and gfm-mode

meain:


raxod502 (commented):
Thanks, however I think it would be best to avoid formatting Markdown code by default, given that the vast majority of Markdown code in the wild is not formatted to Prettier's standards, and most programmers don't expect automated formatting for Markdown. This was previously discussed at https://github.com/radian-software/apheleia/pull/81#pullrequestreview-887880473.

meain:
Woops, somehow missed that when I searched fox existing issues.

raxod502:
No problem, thank you for the work in any case!

PR #147, opened 2022-12-02
Add gofumpt and goimports as formatter options

meain:


raxod502 (commented):
Thanks! Mind adding tests for these formatters, following https://github.com/radian-software/apheleia#adding-a-formatter?

PR #148, opened 2022-12-13
Add support for tsx-ts-mode

danielpza:
Add support for .tsx files (aka react + typescript). `tsx-ts-mode` is one of the new major modes using tree-sitter in emacs 29

from emacs docs:

> tsx-ts-mode is an autoloaded, interactive and natively compiled
function defined in typescript-ts-mode.el.gz.
Signature
(tsx-ts-mode)
Documentation
Major mode for editing TypeScript.
In addition to any hooks its parent mode typescript-ts-base-mode might have run, this mode runs the hook tsx-ts-mode-hook, as the final or penultimate step during initialization.

PR #151, opened 2023-01-25
Add support for remaining treesit modes built into Emacs29

meliache:
This is a continuation of #145, #148. I just used rust-ts-mode from the newly added modes, but then went though NEWS.29 and added whatever else ts-modes I found, assuming the formatter stays the same to the non-treesit-mode.

This PR doesn't cover _all_ ts-modes, there are some remaining modes which previously were not supported in emacs or included in apheleia, like `dockerfile-ts-mode` or `cmake-ts-mode`. There I don't know which formatter to use, if one exists, without doing the research.

The Changelog already covers the addition of new treesit modes, I'll just have to add the PR number once I publish it.


raxod502 (approved):
:sparkles: Thanks!

PR #152, opened 2023-01-31
Avoid overwriting apheleia-formatters for npx commands

stonekyx:
When the formatter is an npx command, setcar is used to expand the
binary file path to node_modules/.bin. But setcar overwrites the
underlying list, so the expanded path is set into
apheleia-formatters. (The list was processed by remq before setcar,
but remq seems to return the cdr if only the first element is
removed.) This is undesirable because when working on multiple NPM
packages, each package may have a formatter of different version.

-------


raxod502 (approved):
:sparkles: Oops, that's embarrassing. Thanks for fixing this.

PR #159, opened 2023-03-09
Fix ktlint

adimit:
Recent versions of ktlint (mine's 0.48.2) have started including logs in their output when the `--stdin` flag is given. These logs would end up being included *into the file* by apheleia, which isn't what we want.

Since emitting log messages is likely never a good idea when formatting code, I just opted to disable them altogether using a command line flag. This fixes the issue.

This is what it looked like:

apheleia

The exact log message (so people can look it up via search) is this:

```
17:35:12.990 [main] INFO com.pinterest.ktlint.internal.KtlintCommandLine - Enable default patterns [**/*.kt, **/*.kts]
```

I've added a `CHANGELOG` entry, even though I'm unsure it's required. I can delete it, if desired.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,12 @@
 All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog].
 
+## Unreleased
+### Bugs fixed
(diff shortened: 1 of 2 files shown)


raxod502:
Thanks!

PR #161, opened 2023-03-12
Separate formatters definitions from core apheleia library

mohkale:
Moves the bulk of apheleia.el into a new apheleia-core.el file and move formatter definitions and mode-associations into a new apheleia-formatters.el file. The end user experience should be unchanged. Loading `apheleia.el` will still load everything load everything, just like it used to.


Makefile
```diff
@@ -52,6 +52,8 @@ checkindent: ## Ensure that indentation is correct
 	    emacs -Q --batch \
 		-l scripts/apheleia-indent.el \
 	        --eval "(setq inhibit-message t)" \
+            --eval "(setq load-path \
+                          (append (list default-directory) load-path))" \
```
raxod502, line 56: I think this needs tabs instead of spaces for Make

raxod502 (approved):
:sparkles:

PR #166, opened 2023-03-12
Add various helper functions for defining formatters

mohkale:
These will be used in the next PR/s I create to add new formatters or update the existing ones to use them.

### DIFF
--- a/apheleia-formatters.el
+++ b/apheleia-formatters.el
@@ -7,6 +7,7 @@
 ;;; Code:
 
 (require 'cl-lib)
+(require 'subr)
 
 (defun apheleia-formatters-indent (tab-flag indent-flag indent-var)
   "Set flag for indentation.
--- a/apheleia-formatters.el
+++ b/apheleia-formatters.el
@@ -2,9 +2,101 @@
 
 ;;; Commentary:
 
-;; Formatter definitions for `apheleia'.
+;; Formatter helper functions and definitions for `apheleia'.
 
 ;;; Code:
+
+(require 'cl-lib)
+
+(defun apheleia-formatters-indent (tab-flag indent-flag indent-var)
+  "Set flag for indentation.
+Helper function for `apheleia-formatters' which allows you to supply
+alternating flags based on the current buffers indent configuration. If the
+buffer is indented with tabs then returns TAB-FLAG. Otherwise if INDENT-VAR
+is set in the buffer return INDENT-FLAG and the value of INDENT-VAR. Use this
+to easily configure the indentation level of a formatter."
+  (cond
+   (indent-tabs-mode tab-flag)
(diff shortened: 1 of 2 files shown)


apheleia-formatters.el
```diff
@@ -2,10 +2,102 @@
...
+  (apheleia-formatters-indent
+   tab-flag indent-flag
+   (cl-case major-mode
+     (json-mode 'js-indent-level)
+     (json-ts-mode 'json-ts-mode-indent-offset)
+     (js-mode 'js-indent-level)
+     (js-jsx-mode 'js-indent-level)
+     (js2-mode 'js2-basic-offset)
+     (js2-jsx-mode 'js2-basic-offset)
+     (js3-mode 'js3-indent-level))))
```
mohkale, line 39: This is kind of an annoying maintenance burden. I don't want to get to the point where we're maintaining the associations between various modes and indent variables :/. I normally use [editorconfig](https://github.com/editorconfig/editorconfig-emacs/blob/e1a391a618ec33d157822dbcc51d010559289f1a/editorconfig.el#L180) for this. If you're okay with adding a dependency on it then I think we'd get much more comprehensive coverage in apheleia through it.

test/formatters/apheleia-ft.el
```diff
@@ -243,41 +247,14 @@ environment variable, defaulting to all formatters."
                 (copy-to-buffer stdout-buffer (point-min) (point-max))))
           (progn
 
+            (let ((result (apheleia--format-command command nil nil)))
+              (setq command (nthcdr 3 result)
+                    in-temp-real-file (nth 0 result)
+                    out-temp-file (nth 1 result)))
```
mohkale, line 252: I'm not sure if there was a reason we didn't use this originally, I found with some of the later formatter updates I did it didn't substitute evaluations that produced nil or lists of values correctly. Rather than continually reverse engineering `apheleia--format-command` I opted to just call it directly :upside_down_face:.

apheleia-formatters.el
raxod502: I think it's perfectly fine to add a dependency on editorconfig if this allows indentation settings to be respected by formatters. Thanks for the idea.

test/formatters/apheleia-ft.el
raxod502: If it works then I'm fine with it. I suspect I didn't realize that the function was sufficiently self-contained and side-effect-free to be usable from tests. Or, it didn't exist when I wrote the test framework. Thanks.

raxod502 (approved):
:sparkles:

apheleia-formatters.el
mohkale: I'll defer this to a follow up.

PR #167, opened 2023-03-12
Update existing formatters to use new formatter helper functions

mohkale:
This follows from #166 and makes the existing formatter definitions in apheleia use the new helper functions. I'd recommend reviewing this after that one is approved/merged.

Note: Looks like emacs defaults to indenting with tabs so most of the test output files have been updated to use tabs over spaces (cue silicon valley meme about tabs > spaces :rofl:).


raxod502:
Maybe it's just me, but is it possible to just make indenting with spaces the default, at least for the test framework? Nobody puts tabs in their json, seriously, no matter what Richard Stallman implemented.

mohkale:
> Maybe it's just me, but is it possible to just make indenting with spaces the default

Easy enough, just updated the tests to default indent-tabs-mode to off.

raxod502 (dismissed):
:sparkles: Wow, this is great. Thank you!

PR #171, opened 2023-03-12
Add buildifier formatter

mohkale:
Follows from #166.


raxod502 (dismissed):
:sparkles: Neat, hadn't even heard of this one before.

PR #175, opened 2023-03-12
Add perltidy formatter

mohkale:
Follows from #166.


raxod502 (approved):
:sparkles: Makes sense.

PR #179, opened 2023-03-17
Take `tab-width` into account in `shfmt`

abougouffa:
The indentation size of 4 was hard-coded in the formatter parameters. This PR fixes this by using the `tab-width` value.


raxod502:
This seems reasonable, however I think the test framework needs some updates to handle using this buffer-local variable in a reasonable way.

mohkale:
#167 should address this in a more accurate way. I think this can be closed.

raxod502:
Yep, looks to be.

PR #188, opened 2023-04-20
utils: Use editorconfig to set formatter indentation

mohkale:


apheleia.el
```diff
@@ -6,9 +6,9 @@
 ;; Created: 7 Jul 2019
 ;; Homepage: https://github.com/raxod502/apheleia
 ;; Keywords: tools
-;; Package-Requires: ((emacs "26"))
+;; Package-Requires: ((emacs "26") (editorconfig "0.9"))
 ;; SPDX-License-Identifier: MIT
-;; Version: 3.2
+;; Version: 3.3
```
raxod502, line 11: I think we only want to bump the version here when we actually release the new version.

PR #198, opened 2023-06-03
Preserve marks as well as points

SohumB:
Marks are processed in almost exactly the same way as point; in particular going through exactly the same `apheleia--align-point` function. This works with minimal changes because markers act like numbers, but also can be passed to `set-marker` to mutate their state.


SohumB:
This is my first nontrivial elisp, so please don't hold back if I've done anything that's not good practice!

SohumB:
Hm. Using this in the past couple of weeks, I've noticed one bit of strange behaviour. In some cases, when I have some marks in a buffer, and then, say, `git revert` large changes to the file, under some conditions I haven't replicated yet, apheleia seems to end up duplicating large chunks of the file; I'm guessing due to trying to match up parts of the old file with parts of the new file. 

I dunno if it has anything to do with trying to preserve multiple marks; I'll switch that around and report back.

raxod502:
Sorry for the delay. I have taken some time away, but am catching up now. Pretty soon I will go through this code and get it merged. Regarding your last comment, I've seen similar things in some rare circumstances; I'll test your branch. Apheleia has also gone too long without any kind of unit test framework that would make it easy to have confidence about making changes like this, so I'll add one of those as well and add tests to cover your change.

SohumB:
Thanks! I just realised I hadn't done the report-back that I mentioned — yes, if I just try to preserve the current mark, that behaviour happens less often but still sometimes. When it occurs, it looks like mark-onwards is duplicated at the end of the file, so at this point I'm reasonably certain we're hitting an edge case of the RCS matching logic.

raxod502:
Brief update: working on the promised test framework, in progress at https://github.com/radian-software/apheleia/pull/204 but not finished yet.

raxod502:
This is taking much, much longer than you should have to wait. Given the delays, as soon as I am able, I will just do a bunch of testing manually, and then merge. Sorry this has taken so long.

raxod502:
Working on resolving merge conflicts, which are my fault for waiting so long to get to this. I think I screwed something up because I get some errors running from the branch now. Will debug and refactor a bit then test.

raxod502:
Did a bit of refactoring here to make the implementation easier to understand.

raxod502:
Needs one or two more bugfixes...

SohumB:
ah woops, sorry, I'd missed all of this! absolutely zero worries about me "having me wait"; I can happily keep using my fork for however long it takes, and I'd much rather you don't feel pressured in the context of an OSS project that I love using so much!

raxod502:
Ok, I think this is working now. Still need better test coverage to be confident with changes like this.

raxod502:
You'll be happy to hear that I have finally implemented a working integration test framework. It was very complex because everything is asynchronous but I sorted out all the issues. Future changes will be able to be tested automatically.

https://github.com/radian-software/apheleia/pull/204

PR #200, opened 2023-07-22
Add support for formatters locally installed via yarn 2+ pnp mode

edslocomb:
This adds support for formatters installed locally in project directories via yarn 2's "zero install" [pnp mode](https://yarnpkg.com/features/pnp).

It's quite similar to the support for formatters installed locally in a project's `node_modules` via npm, and leverages the `npx` symbol, so existing formatter definitions should work without modification.

This checks for a `.pnp.cjs` file (expected in the project root for yarn pnp projects), then looks for a yarn executable, and checks the version of yarn to make sure it supports pnp.  If that works, we just push `"yarn"` onto the front of `command`.

I've only tested this with a locally installed `prettier.js`.  It's very much a works-for-me draft, I'm putting in a PR to make sure this is a workable approach before going any further with it.


zarybnicky:
Trying this out, I see that the `yarn --version` invocation takes more than half a second - which, compared to the actual prettier runtime, takes inexcusably long. (Testing with Yarn 4rc45 and Yarn 3.2.0 both.)

In my experience, the existence of `.pnp.cjs` marks the current project as a Yarn 2+ project pretty much without exception and the version check can be skipped - is your experience different?

edslocomb:
I added the version check because I ran into a problem with emacs invoking an outdated system version of yarn instead of the project local version, but I'm fine with leaving it to the developer to make sure they've got their external toolchain in order :)

apheleia-core.el
```diff
@@ -780,7 +784,7 @@ it's first in the sequence"))
           (unless remote-match
             (error "Formatter uses `file' but process will run on different \
 machine from the machine file is available on"))
-	  (setq stdin nil)
+          (setq stdin nil)
```
raxod502, line 787: Thanks for fixing that indentation.

apheleia-core.el
```diff
@@ -768,7 +768,11 @@ cmd is to be run."
                     "node_modules"
                     project-dir)))))
             (when (file-executable-p binary)
-              (setcar command binary)))))
+              (setcar command binary))))
+        (when (locate-dominating-file default-directory ".pnp.cjs")
+          (if-let (yarn-executable (executable-find "yarn"))
+              (push yarn-executable command)
+            (message "apheleia: .pnp.cjs found but yarn is not installed"))))
```
raxod502, line 775: So, as someone who hasn't used Yarn PNP yet, is the expected usage that you run `yarn prettier` in order to invoke the correct version of Prettier? And `npx prettier` does not work with PNP projects?

edslocomb: Yes, PNP mode does not create a node_modules/ dir, and `yarn foo` is how you run your locally installed executable `foo` using all the locally installed dependencies.

apheleia-core.el
edslocomb: Thank emacs for that ;)

apheleia-core.el
raxod502: If the binary isn't managed by your project, does Yarn fall back to invoking from the rest of `$PATH`? If not, we should add a check to ensure that the binary exists before prepending `yarn`.

edslocomb: I think that's handled by line 773? 
```elisp
(if-let (yarn-executable (executable-find "yarn"))` ?
```
I cribbed the call to `executable-find` from the code that handles npx/node-modules, so I might be misunderstanding what it does..?

raxod502: No, I don't mean check if the `yarn` binary exists, I mean check if the binary the user wants to invoke is installed as part of the current project. If you're in a PNP project that does not have Prettier installed as a dependency, then Apheleia should fall back to invoking a system-global Prettier instead.

edslocomb: Ah!  I see now.

I can think of three ways to address this:

1. Parse the project's package.json file to see if prettier is locally installed.  I'd want to use elisp's json-parse-string for this, which won't work if emacs hasn't been compiled with json support.
2. Use `(shell-exec-to-string (concat yarn-executabnle " info prettier"))`.  This will reintroduce the delay @zarybnicky [noted](https://github.com/radian-software/apheleia/pull/200#issuecomment-1656687846) with the yarn version check.
3. Check for the presence of `.yarn/sdks/prettier/`.  This will not work unless the user has manually installed the base sdks via `yarn dlx @yarnpkg/sdks base`

Thoughts?  Other ideas?

raxod502: I think there is a faster and more reliable way to get the subset of information we are looking for. For example, here is something close to finding the path of prettier:

```
% time node .pnp.js prettier apheleia
[null,"/home/raxod502/.cache/yarn/v6/npm-prettier-3.0.3-432a51f7ba422d1469096c0fdc28e235db8f9643-integrity/node_modules/prettier/index.cjs"]
node .pnp.js prettier apheleia  0.02s user 0.01s system 106% cpu 0.031 total
```

This needs a bit more work to figure out. I will continue looking at this and update here.

raxod502:
Sorry it took so long, but take a look at the latest commits I pushed to your branch. This introduces a new approach which provides full PNP compatibility while adding essentially no overhead (PNP-installed Prettier runs in 148ms with my version, versus 285ms when running via Yarn, versus 115ms when invoking the binary directly).

raxod502:
And CI is broken, because of packaging bullshit. I guess I'll have to fix that too, will do either today or later...

edslocomb:
Excellent!  I was noodling around with a special case of this that would call node directly instead of through yarn, but only for prettier.  The more general approach here is very nice.

I've also pondered spinning up a node process with prettier loaded, and then calling [the api](https://prettier.io/docs/en/api.html) to do the formatting, thus sparing the user the overhead of loading node every time... but that's a project for another day ;)

edslocomb:
Currently working on getting this running in a project with a `.pnp.cjs` and `.pnp.loader.mjs`, not a `.pnp.js`.

raxod502:
Filed https://github.com/prettier/plugin-ruby/issues/1392 upstream about the remaining CI error.

edslocomb:
Getting back to this, with a few notes:

* This won't work in Windows emacs, not sure if there would be affected users
* No idea if it works with tramp mode
* Was puzzled by `apheleia-npx` looking for a `.pnp.js` rather than a `.pnp.cjs`, changed it but should be easy to make it work with both

edslocomb:
Finding I need to add minor fixes to apheleia-ft.el to get CI working; will extract those into a separate PR if the yarn-pnp stuff needs more work after review

raxod502:
Hmm, dunno about .js versus .cjs, when I ran yarn pnp locally it generated .pnp.js so maybe this is a version thing.

pnp-bin/.yarn/cache/@types-emscripten-npm-1.39.8-b45eeffa86-adccdfcec3.zip
raxod502, line 1: Was it intended to include these files?

raxod502 (commented):
Makes sense to me to write this in JavaScript if it is more robust, however having dependencies is not so nice - what leads to the requirement to find formatters in zip files? When would that be usable?

raxod502:
Also, https://github.com/prettier/plugin-ruby/issues/1392 was closed, so applying the suggested fix from there here.

pnp-bin/.yarn/cache/@types-emscripten-npm-1.39.8-b45eeffa86-adccdfcec3.zip
edslocomb: Yes, those zip files are where the dependencies live

edslocomb:
> Hmm, dunno about .js versus .cjs, when I ran yarn pnp locally it generated .pnp.js so maybe this is a version thing.

What does `yarn --version` say?  If I can repo the behavior I'm pretty sure making this work with a .pnp.js will be pretty straightforward

edslocomb:
I've extracted and cleaned up the [pnp-bin](https://github.com/PuddleByteComputing/pnp-bin) js script;  it's set up to build a minified standalone nodejs executable.

The minified standalone is ~290k and targets node 8 or higher.  Would it be better to use that here?

raxod502:
Okay, yeah, that makes sense to me. If we at least have a standalone script, it's still not terribly nice, but at least the vendoring is restricted to a plaintext file. Thank you for doing this! I don't have any other major concerns about the PR.

edslocomb:
Commit [ef82e42](https://github.com/radian-software/apheleia/pull/200/commits/ef82e420b6f985a85d9f15a7e332ada4156df577) should fix https://github.com/radian-software/apheleia/issues/217

raxod502 (approved):
:sparkles:

raxod502:
You think we should merge this for now and then iterate from there? There are getting to be a fair number of issues fixed here but not on main. (I've given you admin permissions on the repository so you can merge things yourself)

edslocomb:
Sounds good to me, there's definitely been some creep in this PR

PR #201, opened 2023-07-23
Add ruby-standard formatter

edslocomb:
There are *so many* opinionated no-config ruby formatters ;)


raxod502 (commented):
Thank you! Changelog entry?

raxod502 (approved):
:sparkles: Thanks!

PR #209, opened 2023-09-17
Use `prettier-json` for `js-json-mode`

cxa:
By default Emacs 29 uses `js-json-mode` for JSON file (correct me if I'm wrong), this pr enables the right formatter for this mode.

### DIFF
--- a/apheleia.el
+++ b/apheleia.el
@@ -250,6 +250,7 @@ rather than using this system."
     (java-mode . google-java-format)
     (java-ts-mode . google-java-format)
     (js3-mode . prettier-javascript)
+    (js-json-mode . prettier-json)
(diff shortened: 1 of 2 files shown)


apheleia.el
```diff
@@ -252,6 +252,7 @@ rather than using this system."
     (js3-mode . prettier-javascript)
     (js-mode . prettier-javascript)
     (js-ts-mode . prettier-javascript)
+    (js-json-mode . prettier-json)
```
mohkale, line 255: This is correct but please sort it so this list stays alphabetical.

raxod502 (commented):
Looks good, changelog update?

raxod502 (approved):
:sparkles:

PR #214, opened 2023-09-30
Remove go-mod-ts-mode from apheleia-mode-alist

akurth:
go-mod-ts-mode is used with go.mod files. These cannot be formatted with gofmt; attempting to do so will lead to an error anyway. Note that also the goimports tool won’t work with go.mod files.


mohkale (approved):
LGTM :-). Thanks.

Issue #181, opened 2023-03-27
Use-package config no longer works after formatter separation


sunng87:
I used to configure apheleia with a  `use-package` section:

```elisp
(use-package apheleia
  :ensure t
  :config
  (setf (alist-get 'rustfmt apheleia-formatters)
        '("rustfmt" "--quiet" "--edition" "2021" "--emit" "stdout"))
  :hook (prog-mode . apheleia-mode))

```

This works until recent refactoring in #161 . The `:config` clause no longer effects. If I eval the `setf` expression after emacs started,  it works. Not sure if it's a bug but I'm curious about how to configure apheleia after the refactoring..

alternateved:
Would that work for you?

```elisp
(use-package apheleia
  :ensure t
  :init
  (require 'apheleia-formatters)
  (setf (alist-get 'rustfmt apheleia-formatters)
        '("rustfmt" "--quiet" "--edition" "2021" "--emit" "stdout"))
  :hook (prog-mode . apheleia-mode))
```

This is what I went with after that update.

sunng87:
As @alternateved pointed out, we have to require apheleia-formatters in before customizing them. 

It turns out the config clause is not evaluated during init. apheleia, at least apheleia-formatters, doesn't seems to load before I open the first `prog-mode`. I get error `(void-variable apheleia-formatters)` when manually eval the `setf` expression. And I believe it's loaded before upgrade because it doesn't require any manual process before.

mohkale:
Well that's surprising an unexpected. To be clear you shouldn't need hacks like what @alternateved shared to work around this. Loading apheleia should load apheleia-formatters and everything else. I can't seem to reproduce the issue using straight instead of the builtin :ensure.

```elisp
(use-package apheleia
  :straight t
  :config
  (setf (alist-get 'rustfmt apheleia-formatters)
        '("rustfmt" "--quiet" "--edition" "2021" "--emit" "stdout"))
  :hook (prog-mode . apheleia-mode))
```

Could someone experiencing the issue help diagnose it. To begin a backtrace of the error from setf and the specific apheleia version (git hash ideally) would be appreciated.

mohkale:
Hmm... I'm not getting an error even from `emacs -q` when running:

```elisp
(require 'package)
(add-to-list 'package-archives '("melpa" . "https://melpa.org/packages/") t)
;; Comment/uncomment this line to enable MELPA Stable if desired.  See `package-archive-priorities`
[...]

PR #215, opened 2023-09-30
Restructure and refactor apheleia package

### LINKED ISSUE #212: Misc fixes
Opening to remind myself to prioritise this:

1. If a formatter is installed on a remote but not locally apheleia will skip formatting.
2. jq formatter currently colorizes the output by default for me.
3. clang-format uses buffer-file-name which doesn't respect remote file paths.

mohkale:
CLOSES #212 

PR to restructure apheleia to make it more modular and improve some of the separation of concerns. Before merging we should make sure that we don't reintroduce the issues from #181.

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -9,7 +9,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        emacs_version: [26, 27, 28, "master"]
+        emacs_version: [27, 28, "master"]
     steps:
       - name: Checkout
         uses: actions/checkout@v2
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,6 +20,8 @@ The format is based on [Keep a Changelog].
 * Prettier is now enabled in `svelte-mode`.
 * More tree-sitter based major modes have been added to
   `apheleia-mode-alist` ([#191]).
+* Refactored the organisation of the apheleia package for ease of
+  understanding and usability ([#215]).
 
 ### Bugs fixed
 * `ktlint` would emit log messages into its stdout when formatting,
@@ -69,6 +71,7 @@ The format is based on [Keep a Changelog].
 [#209]: https://github.com/radian-software/apheleia/pull/209
 [#213]: https://github.com/radian-software/apheleia/pull/213
 [#214]: https://github.com/radian-software/apheleia/pull/214
+[#215]: https://github.com/radian-software/apheleia/pull/215
 
 ## 3.2 (released 2023-02-25)
 ### Features
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -28,6 +28,10 @@ The format is based on [Keep a Changelog].
   and these would get spliced into the source file. This has been fixed
   by suppressing all logs from `ktlint`.
 * Disable colorized output with the jq formatter ([#213]).
+* Fixed apheleia skipped running a formatter on a remote when it isn't
+  installed locally ([#215]).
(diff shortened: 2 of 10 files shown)


raxod502:
Seems like a good idea!

.github/workflows/lint.yml
```diff
@@ -9,7 +9,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        emacs_version: [26, 27, 28, "master"]
+        emacs_version: [27, 28, "master"]
```
mohkale, line 12: Emacs 26 support removed. Looks like that version is so old the executable-find check cannot check remote tramp paths. Rather than adding a workaround I think its fine to just remove it. Supporting almost 3 past version releases seems excessive IMO :thinking:.

apheleia-log.el
```diff
@@ -0,0 +1,105 @@
...
+        (goto-char (point-max))
+        (skip-chars-backward "\n")
+        (delete-region (point) (point-max))
+        (unless (bobp)
+          (insert "\n\n\C-l\n"))
+
+        (unless exit-ok
+          (unless apheleia--last-error-marker
+            (setq apheleia--last-error-marker (make-marker)))
+          (move-marker apheleia--last-error-marker (point)))
```
mohkale, line 73: Possible bug-fix, before the move-marker was within the unless clause. Not sure if that was intentional or not.

mohkale:
Should be good for review now. See each commit for a breakdown of the individual changes. The bulk of this is reorganisation but there are subsequent refactorings and bugfixes etc. No rush to review and merge.

.github/workflows/lint.yml
raxod502: I think that is okay. Emacs 27 is available on the latest Ubuntu LTS release and the latest Ubuntu LTS release has been out for more than a year, so it is fine to drop older versions.

apheleia-log.el
raxod502: Yeah, I think you are right, that is a bug fix and I wrote this incorrectly. The existing code would cause `apheleia-goto-error` to go to the _first_ error, rather than the last one, as documented.

CHANGELOG.md
```diff
@@ -14,18 +14,26 @@ The format is based on [Keep a Changelog].
   respected as before ([#206]).
 * Disable formatting of go module files with gofmt. This was never supported
   ([#214]).
+* Remove support for Emacs 26.0 ([#215]).
```
raxod502, line 17:
```suggestion
* Remove support for Emacs 26 ([#215]).
```

CHANGELOG.md
```diff
@@ -14,18 +14,26 @@ The format is based on [Keep a Changelog].
...
 
 ### Enhancements
 * Use the `prettier-json` formatter for `js-json-mode` ([#209]).
 * Prettier is now enabled in `svelte-mode`.
 * More tree-sitter based major modes have been added to
   `apheleia-mode-alist` ([#191]).
+* Refactored the organisation of the apheleia package for ease of
+  understanding and usability ([#215]).
```
raxod502, line 25: I think this might go under an "internal changes" section.

raxod502 (approved):
:sparkles: I had a look through the commits and all looks good. I like simplifying the number of arguments by moving to defclass - though I've never tried it before in Emacs Lisp, it's worked well for me in other languages.

raxod502:
I'll set `lint (26)` to no longer be a required check since this PR will be removing it. Feel free to merge at your leisure.

Issue #203, opened 2023-08-19
Buffer not being replaced when formatting with lisp function?


nafiz1001:
I've decided to write a lisp function for formatting with spotless because I think the list format wouldn't allow `"-PspotlessIdeHook="` and `filepath` to be concatenated. But when I tried this it would insert the formatted code right below the original code.

```elisp
  (cl-defun spotlessApply+ (&key buffer scratch callback &allow-other-keys)
    (let* ((filepath (buffer-file-name buffer))
	   (process-connection-type nil)
	   (stderr-buffer (generate-new-buffer "*spotlessApply stderr*"))
	   (sentinel (lambda (p e) (progn
				     (funcall callback)
				     (kill-buffer stderr-buffer))))
	   (spotless-process (make-process
			      :name "spotlessApply"
			      :buffer scratch
			      :command `(,(concat default-directory "gradlew")
					 "spotlessApply"
					 ,(concat "-PspotlessIdeHook=" filepath)
					 "-PspotlessIdeHookUseStdOut"
   					 "--quiet")
			      :sentinel sentinel
			      :stderr stderr-buffer)))))
  (push '(spotless . spotlessApply+)
  	apheleia-formatters)

;; temporarily set default-directory to root of project
  (defun shou/fix-apheleia-project-dir (orig-fn &rest args)
    (let ((default-directory (vc-root-dir)))
      (apply orig-fn args)))
  (advice-add 'apheleia-format-buffer :around #'shou/fix-apheleia-project-dir)
```

mohkale:
Note: I plan to fix the inability to concatenate filepath in the work for #212.

raxod502:
The `scratch` buffer starts with the file contents populated in it. You need to modify it in place, or erase it first. (Though if you erase it first then your formatter won't work when chained after another formatter.) If you just insert the formatted contents, you'll end up with two copies, as reported here.

mohkale:
Note: I don't think it's very straightforward to fix the filepath concatenation issue sadly, however you can use `(apheleia-formatters-local-buffer-file-name)` to access the filepath in most situations. There's some nuances around remote files but most of the time that should work fine for you.

nafiz1001:
> Note: I don't think it's very straightforward to fix the filepath concatenation issue sadly, however you can use `(apheleia-formatters-local-buffer-file-name)` to access the filepath in most situations. There's some nuances around remote files but most of the time that should work fine for you.

Thanks your idea worked :)

```elisp
[...]

PR #216, opened 2023-09-30
[#203] Clarify how lisp functions format

raxod502:
Help alleviate confusion like https://github.com/radian-software/apheleia/issues/203.

### DIFF
--- a/apheleia-core.el
+++ b/apheleia-core.el
@@ -906,11 +906,18 @@ being run, for diagnostic purposes."
       ;; with it.
       (insert-buffer-substring (or stdin buffer))
       (funcall func
-               ;; Original buffer being formatted.
+               ;; Original buffer being formatted. This shouldn't be
+               ;; modified. You can use it to check things like the
+               ;; current major mode, or the buffer filename. If you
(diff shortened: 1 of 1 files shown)


apheleia-core.el
```diff
@@ -906,11 +906,18 @@ being run, for diagnostic purposes."
...
+               ;; modified. You can use it to check things like the
+               ;; current major mode, or the buffer filename. If you
+               ;; use it as input for the formatter, your formatter
+               ;; won't work when chained after another formatter.
                :buffer buffer
-               ;; Buffer the formatter should modify.
+               ;; Buffer the formatter should modify. This starts out
+               ;; containing the original file contents, which will be
+               ;; the same a `buffer' except it has already been
+               ;; transformed by any formatters that ran previously.
```
mohkale, line 918:
```suggestion
               ;; Buffer the formatter should modify. This starts out
               ;; containing the original file contents, which will be
               ;; the same as `buffer' except it has already been
               ;; transformed by any formatters that ran previously.
```

mohkale (commented):
Single typo but LGTM

PR #218, opened 2023-10-06
use up-to-date golang

edslocomb:
While running a full `make fmt-build`, I got some failures with installer scripts for golang formatters.  This fixes the problem by installing an up-to-date golang instead of the version available in ubuntu 20.04.

The ppa used here, `longsleep/golang-backports` is [recommended](https://github.com/golang/go/wiki/Ubuntu) by the Go project.


raxod502 (approved):
:sparkles:

PR #221, opened 2023-10-07
Add ormolu and fourmolu, remove brittany

edslocomb:
Found while trying to `make fmt-build` that the installer for `brittany` was broken, and couldn't fix it.

This removes brittany, and adds the ormolu and fourmolu formatters for haskell, and sets ormolu as the default for haskell-mode.


raxod502:
I see, because Brittany is [noted as not currently maintained](https://github.com/lspitzner/brittany/tree/e03ab8425bbc5a3171808cee3285480f64d21536). However, I don't think it's appropriate for us to remove it without notice. Let's leave it in, and fix the installation. I can help with that.

The other additions looks good, though. This PR can just add those two formatters and we can fix Brittany separately.

raxod502 (approved):
:sparkles:

PR #222, opened 2023-10-07
Fix brittany formatter & misc improvements

raxod502:
* Fix brittany installation, needed the `--reorder-goals` argument to be added so that a version that was not the latest version could be installed so that it was compatible with the base lib version shipped with Ubuntu 20.04. We'll upgrade to 22.04 eventually (before it falls out of LTS). Ref: https://github.com/radian-software/apheleia/pull/221
* Add a `make fmt-build-common` target which allows tagging a docker image containing just the base software and not any formatters, to make it easy to debug formatter installation manually.
* Update `apheleia-ft` to also run formatter tests when any files affecting a formatter are changed, which includes the installation script, the sample input/output, and also any scripts (e.g. `apheleia-phpcs`) that it uses. We don't have any logic that will run all formatter tests at once, because that is unwieldy. That can be done manually if making a big change.
* Update to actions/checkout@v4 from v2 because the older one was listed as deprecated.
* Print full stacktraces when `apheleia-ft` fails with an Elisp error.

### DIFF
--- a/.github/workflows/formatters.yml
+++ b/.github/workflows/formatters.yml
@@ -5,7 +5,7 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - name: Checkout pull request
-        uses: actions/checkout@v2
+        uses: actions/checkout@v4
         with:
           ref: ${{ github.event.pull_request.head.sha }}
       - name: Fetch master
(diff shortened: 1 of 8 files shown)


raxod502:
Aha, when running via ssh in GitHub Actions, I get the error printed:

```
fatal: origin/main...HEAD: no merge base
```

No clue what is different that causes it not to be printed in logs normally, must be some kind of buffering.

PR #223, opened 2023-10-10
Use lisp-indent formatter in emacs-lisp-mode

edslocomb:
What it says in the subject :)

Apheleia is opinionated about using a formatter, and which to use as a default, in every other mode it supports, so it seems it should have an opinion about emacs-lisp-mode, too?


raxod502 (commented):
I suppose that's fair. The only thing that sucks is you would not be able to safely edit a file that uses macros which haven't been loaded into Emacs... I suppose we can give it a go and see if people complain. Mind documenting in the changelog?

raxod502 (approved):
:sparkles:

PR #224, opened 2023-10-15
Add ruby-syntax-tree format support

strake7:
Add [ruby-syntax-tree](https://github.com/ruby-syntax-tree/syntax_tree) as an available formatter. Include support for finding a [`.streerc` file](https://github.com/ruby-syntax-tree/syntax_tree#configuration).


strake7:
"Changed formatters" CI failure appears to be unrelated to the changes here. Docker build is failing on install_common.bash. Specifically:
> ...
> #8 37.41 Cannot add PPA: 'ppa:~longsleep/ubuntu/golang-backports'.
> #8 37.41 The user named '~longsleep' has no PPA named 'ubuntu/golang-backports'
> ...

 Assistance with unblocking this would be much appreciated.

test/formatters/installers/ruby-syntax-tree.bash
```diff
@@ -0,0 +1,4 @@
+# Need ruby for gem
+apt-get install -y ruby
+
+gem install syntax_tree --source http://rubygems.org
```
raxod502, line 4: Any reason to specify `http://` explicitly? Or to specify a `--source` at all?

apheleia-formatters.el
```diff
@@ -120,6 +120,8 @@
                 "--stderr" "--format" "quiet" "--fail-level" "fatal"))
     (ruby-standard . ("standardrb" "--stdin" filepath "--fix" "--stderr"
                       "--format" "quiet" "--fail-level" "fatal"))
+    (ruby-syntax-tree . ("stree" "format" filepath
+                         (apheleia-formatters-args-from-file ".streerc")))
```
raxod502, line 124: Is this based on some kind of standard? I would rather avoid Apheleia introducing new configuration files of its own, that won't be compatible with other tooling.

strake7: The configuration file is optional and implemented based on the expectations described by [the Syntax Tree docs](https://github.com/ruby-syntax-tree/syntax_tree#configuration). I understand your concern though. Let me explore the possibility of removing the argument extraction from apheleia and instead provide a path to the rc file.

raxod502: Ah, I see, it is indeed a standard. We should not parse it ourselves, but we should make sure we run from the right working directory. Maybe the same fix as https://github.com/radian-software/apheleia/pull/239?

raxod502:
@strake7 How's the latest commit look to you? I refactored the existing `apheleia-mix` example into a reusable script, and adapted it for use here as well. It will automatically run stree from the same directory as `.streerc` if such a file exists. Then based on the documentation you linked I think it would load the file by default. If there is no such file it will just execute from the working directory.

strake7:
> @strake7 How's the latest commit look to you? I refactored the existing `apheleia-mix` example into a reusable script, and adapted it for use here as well. It will automatically run stree from the same directory as `.streerc` if such a file exists. Then based on the documentation you linked I think it would load the file by default. If there is no such file it will just execute from the working directory.

This works great and is a far more desirable solution. Thank you @raxod502. Closing this PR

raxod502:
Er, well, I didn't merge anything, I just pushed to your PR branch.

test/formatters/installers/ruby-syntax-tree.bash
raxod502, line 4:
```suggestion
gem install syntax_tree
```

raxod502 (approved):
:sparkles:

strake7:
I mixed up my remotes then. @raxod502 An unused function, `apheleia-formatters-args-from-file`, made it in through. I removed it here - #273.

PR #230, opened 2023-10-22
feat: add filepath to mix-format

cepehang:
Allows mix to format files other than elixir or erlang files, e.g. `.heex` files given : 
- [proper formatting directory configuration](https://github.com/radian-software/apheleia/issues/30#issuecomment-778150037)
- [proper mix formatting config](https://hexdocs.pm/phoenix_live_view/Phoenix.LiveView.HTMLFormatter.html#module-setup)

### DIFF
--- a/apheleia-formatters.el
+++ b/apheleia-formatters.el
@@ -72,7 +72,7 @@
     (lisp-indent . apheleia-indent-lisp-buffer)
     (ktlint . ("ktlint" "--log-level=none" "--stdin" "-F" "-"))
     (latexindent . ("latexindent" "--logfile=/dev/null"))
-    (mix-format . ("mix" "format" "-"))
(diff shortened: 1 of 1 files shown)


cepehang:
The `--stdin-filename` option is available since mix version 1.14 but Ubuntu 20 only has mix version 1.9 in its repositories.

raxod502:
One option would be to add an `apheleia-mix-format` script, similar to https://github.com/radian-software/apheleia/blob/main/scripts/formatters/apheleia-phpcs, which would run `mix-format --version` (or equivalent) and cache the output on disk somewhere in `$XDG_CACHE_HOME`, invalidating that cache when the mtime of the `mix-format` binary is updated. Then it would either pass the argument or not, depending on the version.

That might be more effort than it's worth, though, given that you also have the opportunity to just change the formatter options directly in your `init.el`.

PR #231, opened 2023-10-23
Use hclfmt to format hashicorp HCL files.

dbalan:
Add formatter `hclfmt` for Hashicorp Configuration Langauge (HCL).


raxod502 (approved):
:sparkles: Thanks!

Issue #226, opened 2023-10-18
Apheleia sometimes overwrites changes in my JS buffers 


DinoChiesa:
Short summary: when editing JS files, I often move around pretty quickly, edit or add some text, then save, then move again, add some other code, and save. All within a handful of seconds. The "save often" thing is a sort of muscle memory /  mindless habit that happens when I'm "in the zone".  I don't really think about it. 

Now that I'm using Aphelia, If I wait a little while after the second save, often (maybe it depends on timing?) the second set of changes disappears from my buffer. This is in emacs 29.1 .  If I'm not paying attention, I don't even realize it.  I just keep going.  Later, when I run tests, and see confounding results,  I go examine the code and see, the code I had put in place earlier is now gone.  

Demonstration: https://www.youtube.com/watch?v=sppRYlaqquU

The setup for my formatter for my JS mode looks like this: 
```
(push '(my-prettier-js .
        ("/usr/local/bin/prettier"
         "--stdin-filepath" filepath "--parser=babel-flow"
         "--trailing-comma" "none"
         (apheleia-formatters-js-indent "--use-tabs" "--tab-width")))
      apheleia-formatters)

(push '(js-mode . my-prettier-js) apheleia-mode-alist)
```

I think the way I can avoid this is to wait 5-6 seconds after every save, before making any further changes. That feels a little klunky to me. I could also just save ONCE, instead of lots of times.  That's a personal habit that's ingrained, and hard to change.

The README says that Apheleia is designed to not apply the reformatting, _if the buffer contents have changed while running the formatter_.  But I am not seeing that. 

I use Apheleia for various languages (sh, go, java, ...) , and JS is the only case where I've noticed this.  Obviously, the other languages use other formatters. 

Can you make suggestions?  Have I done something wrong with setting up the formatter?  how can I diagnose or debug this?  Or better, how can I solve it and eliminate this behavior?

I tested the prettier binary; when I run it outside of emacs, it seems to run quickly and correctly. Results get displayed in much less than a second.  I haven't looked at how long it takes when run within emacs by apheleia in the after-save-hook.  Not sure this is the problem, especially since Apheleia is supposed to not apply re-formatting if the file has changed. If that's so, it shouldn't matter how long the formatter takes to run, right? 

I really like Apheleia, thanks for writing it,  it's mostly really good. But it edges into chaotic evil territory when some of my changes get lost, some of the time!

Thanks

DinoChiesa:
[...]

PR #234, opened 2023-10-29
[#226] Use lexical variable for saved buffer hash

raxod502:
Should help with https://github.com/radian-software/apheleia/issues/226 although there is a second problem related to performance also reported in that thread, which would not be helped with this.

Mostly indentation changes.

### DIFF
--- a/apheleia.el
+++ b/apheleia.el
@@ -44,9 +44,6 @@
       (buffer-hash)
     (md5 (current-buffer))))
 
-(defvar apheleia--buffer-hash nil
-  "Return value of `buffer-hash' when formatter started running.")
-
 (defun apheleia--disallowed-p ()
   "Return an error message if Apheleia cannot be run, else nil."
   (when (and buffer-file-name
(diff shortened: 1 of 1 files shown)


raxod502:
This uses the lexical variable approach suggested by @DinoChiesa, which I think is the best solution. There is a second thing mentioned in https://github.com/radian-software/apheleia/issues/226#issuecomment-1769166750, which is that we should perhaps only allow the results of the _last_ (most recent) formatting operation to be applied, rather than whichever one happens to finish first. I think this implementation should actually achieve that as a side effect.

In the following cases the "earlier formatter" and "second formatter" are two instances of the same formatter. Chained formatters are configured to run in series, and that is not what is happening in #226.

Case 1: no changes to buffer between when the two format operations start. The formatter makes no changes to the buffer. In this case it doesn't matter which order the formatters finish in because neither will make any changes.

Case 2: no changes to buffer between when the two format operations start. The formatter wants to make some changes to the buffer. If the earlier formatter finishes first then its changes will invalidate the saved buffer hash from the second formatter so the second formatter's changes will not be applied. The same happens if the later formatter finishes first. Either way only one formatter runs, and since there were no changes to the buffer between when the two format operations start, the output should be the same in both cases.

Case 3: the buffer was modified between when the two format operations start. If the earlier formatter finishes first, it will abort because the buffer hash changed. The second formatter will be able to apply its changes (or it won't have any to apply) because the buffer hash has not changed since it started. If the later formatter finishes first, the same thing will happen. The earlier formatter will still abort as before... unless the buffer modification is actually reverted by the later formatter which finished first. In that case the earlier formatter will indeed get to run as well. However that should be fine because it will have the same effect as running the same formatter twice, no data will be lost.

Issue #233, opened 2023-10-29
Add support for `ruff format` for Python files


offbyone:
The `ruff` linter has gained formatter support: https://github.com/astral-sh/ruff#usage

It's in beta, but it's mostly black-compatible, but much faster. I'd love it if Python formatting could use the "right" formatter depending on if `black` or `ruff` are installed. 

I'm happy to submit a PR for this support.

raxod502:
Sure, that sounds fine to me. Let's keep `black` as the default for now, but add support for `ruff` so that users can set it to be used for Python if they prefer.

offbyone:
I have opened a PR for this; I'm happy to add any changes you see fit, but it's at least passing checks :D

PR #236, opened 2023-10-30
Add support for the ruff python formatter

offbyone:
This adds support, but does not make it the default
Fixes: #233


raxod502 (commented):
Looks good to me, mind adding a changelog entry?

raxod502 (approved):
:sparkles:

PR #237, opened 2023-11-04
Fixes for test installers

edslocomb:
Fix various installers, to get a clean `make fmt-build`

1. Backports from #200 
  - Use new universal installer for Nodejs
  - install prettier-ruby and prettier-svelte plugins locally in /tmp
  - use newer prettier option `--plugin=` instead of `--parser=` for
  prettier-ruby and prettier-svelte
2. Use standard local rust/rustfmt install for test setup instead of tarball (no longer available)
  - install locally, change permissions to make the install visible to all users, and then symlink 
  to the install dir and config in the global /etc/bash.bashrc so local test user will picks them up.
3. Change filename to download for nixfmt (now includes OS/arch)
  - There was initially no binary available when nixfmt was updated,
  but the devs there were kind enough to provide one: 
  https://github.com/serokell/nixfmt/issues/139


raxod502 (dismissed):
:sparkles:

raxod502:
We might also need to bump the base image version, it's currently 20.04 but that ships Emacs 26, which https://github.com/radian-software/apheleia/pull/215 dropped support for.

edslocomb:
Closed in favor of #242

edslocomb:
Closed in favor of #242

PR #239, opened 2023-11-04
[#232] Format from elixir project dir

### LINKED ISSUE #232: Elixir formatter not using project `.formatter.exs` file
When formatting a file within a project that has a `.formatter.exs` file, apheleia is not using this config to format the file.

raxod502:
Make it so that if there is a `.formatter.exs` file somewhere in the parent directories, then Apheleia will run `mix format` from there instead of the current directory.

Close #232


raxod502:
Seems like there are some CI issues which I'll have to fix

raxod502:
Nice, @edslocomb fixed CI for me in https://github.com/radian-software/apheleia/pull/242 :)

PR #242, opened 2023-11-06
Bump test environment to Ubuntu 22.04

edslocomb:
This also fixes a few test installers:

- rustfmt is available via apt-get in 22.04, use that instead of tarball that's no longer available
- brittany now installs cleanly in the default haskell ghc/cabal in 22.04, so just use that
- perltidy in 22.04 has slightly different output
- nixfmt download url changed (now includes OS/arch). There was initially no binary available when nixfmt was updated,
  but the devs there were kind enough to provide one: https://github.com/serokell/nixfmt/issues/139

This runs a full `make fmt-build` without errors, though it takes a while ;).  All tests pass, too.


raxod502 (approved):
:sparkles: Yay!

PR #243, opened 2023-11-07
move pnp-bin.js into formatters dir for simpler melpa recipe

edslocomb:
Gets all non-elisp dependencies into a single directory, `scripts/formatters`


edslocomb:
Apologies for merging this (small) change without review, I want to get a PR into melpa updating the recipe to include the non-elisp support files (see https://github.com/radian-software/apheleia/issues/241)

raxod502:
No worries, you don't need to get review on things unless you feel it's needed.

PR #249, opened 2023-11-20
Add support for `dprint`

scop:
https://dprint.dev


apheleia-formatters.el
```diff
@@ -44,6 +44,7 @@
     (cmake-format . ("cmake-format" "-"))
     (crystal-tool-format . ("crystal" "tool" "format" "-"))
     (dart-format . ("dart" "format"))
+    (dprint . ("dprint" "fmt" "--stdin" (file-name-nondirectory filepath)))
```
scop, line 47: My elisp is rusty and this fails:
```
Debugger entered--Lisp error: (void-variable filepath)
  (file-name-nondirectory filepath)
```

We need to pass a file basename only (or extension only), dprint treats full paths specially; in particular they must exist which causes problems e.g. with the test suite at the moment, and I suspect it might in other scenarios, too.

apheleia-formatters.el
```diff
@@ -239,11 +240,13 @@ rather than using this system."
     (cmake-mode . cmake-format)
     (cmake-ts-mode . cmake-format)
     (common-lisp-mode . lisp-indent)
+    (conf-toml-mode . dprint)
```
scop, line 243: dprint supports lots of things. I opted to add here only some of the most common associations that did not have any yet: toml, Dockerfile, markdown.

dprint also requires one to have a plugin installed for each type, so there's no guarantee that even these would work -- there might not be a plugin installed for these or a config in effect might not reference one, etc. Perhaps the safer bet would be to not add any here :thinking:

test/formatters/installers/dprint.bash
```diff
@@ -0,0 +1,11 @@
...
+cat <<\EOF >/tmp/dprint.json
+{
+  "toml": {},
+  "excludes": [],
+  "plugins": [
+    "https://plugins.dprint.dev/toml-0.5.4.wasm"
+  ]
+}
+EOF
+dprint config update
```
scop, line 11: This feels dirty, but gets things going. As far as the tests are run in a subdir of /tmp somewhere -- dprint searches config from current dir and ancestors.

We could add a bunch of other plugins and test files if wanted. Can add if so, once the `file-name-nondirectory` issue above is solved.

apheleia-formatters.el
raxod502: I think what we want is probably some way to communicate to Apheleia whether or not the plugin is installed. Can that be checked quickly? Currently, Apheleia checks whether the binary to be executed for a formatter is on `$PATH`, and if it is not, then it silently will skip formatting, because of course trying to invoke it would fail. We could add a syntax that could be used to specify a custom check for whether the formatter exists.

Or, alternatively, we could use the existing facility of adding wrapper scripts like `apheleia-phpcs` and `apheleia-npx`, and pass the plugin name as an argument. The wrapper script would check its existence, silently return if the plugin is not available, and delegate to it otherwise.

apheleia-formatters.el
raxod502: It's not your fault. You're thinking that `filepath` is an ordinary variable, which would make sense. But it's actually a special symbol that is replaced before the command line is evaluated. The replacement only happens at the top level.

I think it would be reasonable to bind `filepath` and related symbols when evaluating non-string forms as mentioned in the docstring in `apheleia-formatters`. Then what you did would work.

By the way, I think you found a bug in the test framework. I think whatever `filepath` expands to _should_ exist during test execution. There might have been an oversight where that is not the case currently.

test/formatters/installers/dprint.bash
raxod502: Yeah, multiple `in.*` / `out.*` pairs is supposed to work. See the `prettier` subdir for example.

Agree that this doesn't look great, but I think the real answer is there needs to be a more mature test framework that would allow the inclusion of extra files to be placed on the filesystem during test invocation for a particular formatter, but it is okay to not bother with that for now since this works.

apheleia-formatters.el
```diff
@@ -269,6 +272,7 @@ rather than using this system."
     (LaTeX-mode . latexindent)
     (lua-mode . stylua)
     (lisp-mode . lisp-indent)
+    (markdown-mode . dprint)
```
raxod502, line 275: I've been holding off on installing a default formatter for markdown because people have so many different ways of formatting it across different projects. Will add a comment noting as such so it's clear why there is seemingly an entry missing.

apheleia-formatters.el
scop: I'm not sure if that plugin existence can be checked that quickly/cleanly. Would need to traverse dirs upwards, locating one of the 4 different allowed filenames (following dprint's order of precedence), and then grabbing something out of the found file which can be json or jsonc, https://dprint.dev/setup/

But I found that dprint falls back to just passthrough without any formatting (like `cat`) if it gets passed a filename that is not handled by the installed plugins. I think with this behavior we could just add whatever associations we see appropriate and know are working with _some_ dprint+plugin config to the default `apheleia-mode-alist` -- without an applicable plugin, dprint wouldn't just change anything. I'll leave `config-toml-mode` and `dockerfile-mode` there for now, but I'm all good if we'd rather not add any dprint entries there for now.

apheleia-formatters.el
scop: Ok, good, removed in 7074bec0c0884a23a7ba959b61026ff18432b069.

(I'm assuming you'll eventually squash this PR on merge, so keeping the branch history here more reviewable than tidy.)

test/formatters/installers/dprint.bash
scop: Ok, added some more plugins and sample code in e13e245dcce7acd8561f2cf28a32db1a9419d43b.

apheleia-formatters.el
raxod502: I looked into this. The input file buffer has its buffer name set to an absolute filepath (unusual):

https://github.com/scop/apheleia/blob/cf05ccf000c66478151cc9d5897ac8597f0b68fc/test/formatters/apheleia-ft.el#L295

The `filepath` symbol is replaced here:

https://github.com/scop/apheleia/blob/cf05ccf000c66478151cc9d5897ac8597f0b68fc/apheleia-formatters.el#L894-L898

Because the buffer name is not supposed to be critically important it is sanitized:

https://github.com/scop/apheleia/blob/cf05ccf000c66478151cc9d5897ac8597f0b68fc/apheleia-formatters.el#L810-L817

That explains the unusual path name showing up. The bug is that the buffer name is used at all. Obviously, a buffer name does not a `filepath` make. The bug is that the test framework is not setting `buffer-file-name` in the test buffer, which is causing Apheleia to fall back to a non-existent `filepath` based on the buffer name.

I will fix this.

apheleia-formatters.el
```diff
@@ -47,7 +47,7 @@
                   (apheleia-formatters-indent
                    "--indent-with-tabs" "--indent-size"))
     (dart-format . ("dart" "format"))
-    (dprint . ("dprint" "fmt" "--stdin" (file-name-nondirectory filepath)))
+    (dprint . ("dprint" "fmt" "--stdin" filepath))
```
raxod502, line 50: As discussed - it should be fine to give the full `filepath`, with the other fixes below.

test/formatters/apheleia-ft.el
```diff
@@ -269,8 +269,10 @@ environment variable, defaulting to all formatters."
     (dolist (in-file (apheleia-ft--input-files formatter))
       (let ((extension (file-name-extension in-file))
             (in-text (apheleia-ft--read-file in-file))
+            ;; The `in-temp-real-file' variable is set to whatever
+            ;; temporary file the formatter will run on (in case it
+            ;; uses the `file' or `filepath' symbol or is a function).
```
raxod502, line 274: It took me a while to understand the purpose of this variable. Adding a comment to explain the difference from `in-temp-file`, which also doesn't actually have to be defined up here because it is only used temporarily and then discarded.

test/formatters/apheleia-ft.el
```diff
@@ -292,30 +294,36 @@ environment variable, defaulting to all formatters."
         ;; Some formatters use the current file-name or buffer-name to interpret the
         ;; type of file that is being formatted. Some may not be able to determine
         ;; this from the contents of the file so we set this to force it.
-        (rename-buffer in-file)
+        (rename-buffer (file-name-nondirectory in-file))
```
raxod502, line 297: Use `file-name-nondirectory` would have made the behavior you saw less confusing - it's unusual to set buffer names to full filesystem paths. Normally they are just basenames.

test/formatters/apheleia-ft.el
```diff
@@ -292,30 +294,36 @@ environment variable, defaulting to all formatters."
...
                              (format "*apheleia-ft-stdout-%S%s" formatter extension)))
         (with-current-buffer stdout-buffer
           (erase-buffer))
         (if (functionp command)
-            (progn
-              (setq in-temp-file (apheleia-ft--write-temp-file
-                                  in-text extension))
+            (let ((in-temp-file (apheleia-ft--write-temp-file
+                                 in-text extension)))
+              (setq in-temp-real-file in-temp-file)
```
raxod502, line 305: This should have been setting `in-temp-real-file`, which is checked later to verify that it wasn't modified by the formatter writing to the input file. Setting `in-temp-file` was doing nothing because that variable isn't ever read again.

test/formatters/apheleia-ft.el
```diff
@@ -292,30 +294,36 @@ environment variable, defaulting to all formatters."
...
+                          (intern formatter) command nil nil)))
+                (setq command `(,(apheleia-formatter--arg1 ctx)
+                                ,@(apheleia-formatter--argv ctx))
+                      ;; In this case the real temp file might be
+                      ;; different from the one we generated, because
+                      ;; the context creator might generate another
+                      ;; temporary file to avoid touching our existing
+                      ;; one.
+                      in-temp-real-file (apheleia-formatter--input-fname ctx)
+                      out-temp-file (apheleia-formatter--output-fname ctx))))
```
raxod502, line 326: The core of the bug here was failing to run `apheleia--formatter-context` in a buffer corresponding to the input file, which meant it did not have a `buffer-file-name` set. The buffer was right in the other code path for handling functional formatters.

test/formatters/samplecode/dprint/out.ts
```diff
@@ -0,0 +1,6 @@
+interface GreetingSettings {
+  greeting: string;
+  duration?: number;
+  color?: string;
+}
+declare function greet(setting: GreetingSettings): void;
```
raxod502, line 6: This output file was missing.

raxod502:
This might need one or two more changes on my part, I think the `rename-buffer` call is also in the wrong context, and we should be hoisting the creation of the temporary input file to the beginning and not having it show up in both conditional branches. The entire test should probably be run in the context of the input temporary file, it was confusing me greatly that that was not the case.

raxod502:
Let's merge and fix forward

raxod502 (approved):
:sparkles:

PR #250, opened 2023-11-21
Add support for `yq`

scop:


raxod502 (approved):
:sparkles: Neat! I actually didn't know yq supported that many formats.

PR #252, opened 2023-11-21
Respect `typescript-indent-level`, `typescript-ts-mode-indent-offset`

andersk:
https://github.com/emacs-typescript/typescript.el/blob/v0.4/typescript-mode.el#L635-L639
https://github.com/emacs-mirror/emacs/blob/emacs-29.1/lisp/progmodes/typescript-ts-mode.el#L38-L43


raxod502 (approved):
:sparkles:

PR #253, opened 2023-11-22
Always supply `--stdin-filepath` to Prettier

andersk:
This is required for Prettier to pick up the correct settings from its config files.


raxod502 (commented):
Thank you! Do you mind adding a note to the changelog about this bugfix?

andersk:
Added.

raxod502 (approved):
:sparkles:

PR #257, opened 2023-11-24
Add support for JSON with `python3 -m json.tool`

scop:
https://docs.python.org/3/library/json.html#module-json.tool


raxod502 (approved):
:sparkles: Looks good to me!

PR #258, opened 2023-11-25
Determine indent level from more modes

scop:
Pass CSS, GraphQL, HTML, Ruby, Svelte, and YAML mode indent setting to formatters that make use of it through `apheleia-formatters-js-indent`.

`prettier-markdown` and `prettier-scss` attempt to make use of this too, but `markdown-mode` and `scss-mode` don't have such a setting that I can tell. Left the uses in place for `indent-tabs-mode` support anyway.

BTW the `apheleia-formatters-js-indent` name and its description does not seem quite spot on, as it was already being (partially unsuccessfully) (ab?)used by some modes for which "JavaScript like"ness is a bit of a stretch :)


scop:
Reworked against current main branch.

raxod502 (dismissed):
:sparkles: Thank you! Sorry for merge conflicts.

raxod502:
At least we got that pesky variable renamed though with the refactor!

raxod502 (approved):
:sparkles:

PR #263, opened 2023-11-26
Add support for Robot Framework files with `robotidy`

scop:


apheleia-formatters.el
```diff
@@ -136,6 +136,9 @@
         "--parser=yaml"
         (apheleia-formatters-js-indent "--use-tabs" "--tab-width")))
     (purs-tidy . ("apheleia-npx" "purs-tidy" "format"))
+    (robotidy . ("robotidy" "--no-color" inplace
```
scop, line 139: I wonder if this triggers a bug somewhere, test suite or `inplace` implementation, or my understanding of it?
```
Debugger entered--Lisp error: (error "Formatter robotidy modified original file in place")
``

Comment (mohkale) [reply apheleia-formatters.el:139]: Ouch. This is probably a bug. It's the first time we've had a formatter actually use inplace :upside_down_face:. Could you open a new issue for this? 

In the meantime I took a look at robotidy and it supports formatting to stdout. Generally this is the better choice because it side-steps the filesystem for subprocess io. You can change inplace to "-" and it should work.

File: CHANGELOG.md [hunk 2]
```diff
@@ -3,6 +3,11 @@
 All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog].
 
+## Unreleased
+
+### Formatters
+* [`robotidy`](https://robotidy.readthedocs.io) for Robot Framework files.
```
mohkale, line 9: Please add a link to this PR as well.

apheleia-formatters.el
scop: Opened https://github.com/radian-software/apheleia/issues/265 for the test suite issue.

Re stdout, nice catch! I couldn't find anything related to https://robotidy.readthedocs.io/en/stable/quickstart.html#transform-code-from-standard-inputthat in the docs, but now I see https://robotidy.readthedocs.io/en/stable/quickstart.html#transform-code-from-standard-input (that refers to stdin only though...). Anyway it seems to work indeed.

CHANGELOG.md
scop: Done.

mohkale (approved):
LGTM. Thanks.

PR #264, opened 2023-11-27
Add denofmt

dalugm:
I havn't checked my PR for a very long time, sorry for the long delay...

As for now, I think it's better to follow convention, so I closed #142 and open a new. And I'd like to use deno more since sometimes I don't have a node but deno...


mohkale (commented):
LGTM. Thanks. Please add a changelog entry and then we can merge.

dalugm:
> LGTM. Thanks. Please add a changelog entry and then we can merge.

Sorry I forgot to add it, it should be fine now

mohkale (approved):
Thanks a bunch :sunglasses:.

Issue #266, opened 2023-11-29
How to add formatter that always returns positive exit code on changes (e.g. `docformatter)


meliache:
I tried setting up [docformatter](https://docformatter.readthedocs.io/en/latest/usage.html) as a formatter for formatting python docstrings. By default it outputs a diff but with `--in-place` it fixes the file.

But it didn't work for me because in both modes of operation returns a non-zero error code when it found a style mistake, even if it successfully fixed it with `--in-place`. Can I somehow suppress the error. I tried adding to `apheleia-formatters` the element
``` elisp
(docformatter "docformatter" "--in-place" inplace "||" "true")
```
But this `||` didn't work due to shell escaping in Emacs and probably `make-process` doesn't allow for pipes anyway.

Anyway, I my temorary solution was just to write a bash script wrapper that adds docformatter and suppresses errors, but that solution isn't nice...

raxod502:
We do this for the phpcs formatter:

https://github.com/radian-software/apheleia/blob/8f512dba3ee2eabae0dbfd8289b89d54c132cb87/scripts/formatters/apheleia-phpcs#L1-L6

If you add a script to that directory then it will be automatically available on Apheleia's `$PATH`.

The advantage of this approach is that you can identify the specific nonzero exit codes that are to be ignored, but other presumably unexpected errors such as interrupts will not be treated as success.

PR #267, opened 2023-12-04
Add `docformatter` which formats Python docstrings to PEP 257

meliache:
Add [docformatter](https://github.com/PyCQA/docformatter) for Python docstrings.

By default it outputs diffs but changes in-place with `--in-place`. On successful change it exits with an error code of `3` (found out by trial), so I had to add a formatter wrapping-script.

Initially I used `--in-place`  with the special `in-place` symbol in apheleia. But now I tried an approach where I transform the diff into usable stdout using `patch` instead.

Related to #266 , where I had used the example of docformatter to ask how to add scripts with positive exit codes and @raxod502 showed me the `phpcs` solution.


meliache:
I have possible implementations of a docformatter wrapper script and both work interactively but fail the tests. I think I need some help. maybe @raxod502 has some idea?

### Implementation 1: Using `--in-place`

Script:
``` shell
#!/bin/sh
docformatter --in-place "$@"
if [ "$?" -eq 3 ]; then
	exit 0
fi
```
Formatter:
``` elisp
    (docformatter . ("apheleia-docformatter" inplace))
```
Error: 
```
  error("Formatter %s modified original file in place" "docformatter")
  ```
  In place is exactly what I tried to do here, supposedly apheleia has support for that, but seems it doesn't work for tests.  
  
### Implementation 1: Transformind diff to stdout

Script:
``` shell
#!/bin/sh
output=$(docformatter "$@")
ret=$?
[ $ret -gt 0 ] && [ $ret -ne 3 ] && exit $ret
echo "${output}" | patch --quiet --output=-
```
Formatter:
``` elisp
    (docformatter . ("apheleia-docformatter" file))
```
Error: 
```
[Errno 2] No such file or directory: '/tmp/srctestformatterssamplecodedocformatterin.py'
  ```
  
When testing interactively after a while I found the call to `patch` with standard input is very sensitive to from where it's called. It works when it's called in the same directory as the `in.py` but when it's in another directory it doesn't work except with a correct `--strip` argument. Not sure if this is what causes this weird error message.

raxod502:
It seems simpler to go with in-place in this case, and the feature is supported for a reason. Let me have a look and see if I can fix the issue with the test framework.

raxod502 (approved):
:sparkles:

meliache:
Thanks for fixing and merging this! I was positively surprised that this package even has a dockerized testing framework in the first place, I have contributed to enough packages with no unit tests and have no personal experience with doing them in Elisp.

PR #269, opened 2023-12-06
Log an error when a formatter's executable is not found

### LINKED ISSUE #268: No feedback/logging when formatter command is not found in the exec-path
Hi! Thanks for apheleia, it is really cool! I was trying to add an entry for the [Ruby syntax_tree](https://github.com/ruby-syntax-tree/syntax_tree) formatter. I wasn't getting any output and wasn't sure what was happening, even after setting `apheleia-log-debug-info` to `t`. I would get a log like this:

```
<format-buffer>: Invoking apheleia-format-buffer on #<buffer filepath.rb> with formatter (stree)
<run-formatter>: Running formatters (stree) on buffer #<buffer filepath.rb>
```

but nothing after this.

After looking at https://github.com/radian-software/apheleia/blob/8f512dba3ee2eabae0dbfd8289b89d54c132cb87/apheleia-formatters.el#L959 for a bit and trying out a formatter which worked, I realized that it tries to look up the executable in the exec-path but it seems like if it doesn't find the executable it doesn't log anything. I think it'd be nice if we could log _something_ in this case so that people know what's wrong. In my case my exec-path was set to something custom for some reason (when I unset it, it changed to match my shell $PATH which is really all I wanted).

My elisp is a bit rusty but maybe I can contribute a patch to log something when this happens... in any case, I wanted to at least log a note of this somewhere in case someone else runs into this.

I'd also like to submit a patch to add support for the syntax_tree gem, although for those using prettier it's essentially the same as the prettier-ruby formatter. But I wanted to have it working for all projects and not just those that are using Prettier/JS packages so I wanted to have an entry for syntax_tree standalone.

ibrahima:
I was trying to set up a new formatter and it wasn't in my exec-path, so it wasn't running anything. But there was no message in the apheleia logs, even with `apheleia-log-debug-info` enabled. This adds a log entry when the executable is not found.

Hopefully the formatting is okay - I think Apheleia re-indented it so hopefully that's good.

Fixes #268

### DIFF
--- a/apheleia-formatters.el
+++ b/apheleia-formatters.el
@@ -980,26 +980,25 @@ purposes."
                               (let ((load-suffixes '(".el")))
                                 (locate-library "apheleia"))))))
                         exec-path)))
-      (unless
-          (when (executable-find (apheleia-formatter--arg1 ctx)
-                                 (eq apheleia-remote-algorithm 'remote))
-            (apheleia--execute-formatter-process
-             :ctx ctx
-             :callback
-             (lambda (stdout)
-               (when-let
-                   ((output-fname (apheleia-formatter--output-fname ctx)))
-                 ;; Load output-fname contents into the stdout buffer.
-                 (with-current-buffer stdout
-                   (erase-buffer)
-                   (insert-file-contents-literally output-fname)))
-               (funcall callback stdout))
-             :ensure
-             (lambda ()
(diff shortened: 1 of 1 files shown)


raxod502 (approved):
It's a little more typical to use `if` for this, I think. Updated. Otherwise, looks good, thank you! The logging facility is new and can stand improvement, for sure.

PR #275, opened 2023-12-28
Extend lua support to lua-ts-mode

dpassen:
Available via https://git.sr.ht/~johnmuhl/lua-ts-mode and https://git.savannah.gnu.org/cgit/emacs.git/commit/?id=b6659e98a4fcaa44477b64d7782243feca020418


raxod502 (approved):
:sparkles:

PR #279, opened 2024-01-04
Add `ruff-isort` formatter

offbyone:
`ruff` doesn't run isort as part of its format, so to maintain parity with the sort rules that black/isort offer, this is the ruff isort.


raxod502 (commented):
```
warning: Ignoring file chech in favor of standard input.
```

I was _wondering_ how that could work... let me fix

raxod502 (approved):
:sparkles:

offbyone:
Woot, thanks!

PR #280, opened 2024-01-08
Add support for `treefmt`

terlar:
This formatter is formatting via [treefmt](https://numtide.github.io/treefmt). This is useful if the project is configuring formatters for the project and you don't have to replicate this logic within Emacs, but instead utilize treefmt directly.

To do that you can set the `apheleia-formatter` to `treefmt` within the projects `.dir-locals.el`.

I noticed there are tests for this, since treefmt could technically format anything, I went with treefmt formatting rust code. It also relies on having a `treefmt.toml` in the project folder, so had to add that as well in the tests.


raxod502 (approved):
:sparkles:

PR #282, opened 2024-01-26
add support for pyang

sgherdao:
Hi, 

pyang (https://github.com/mbj4668/pyang) is a YANG validator, transformator and code generator, written in python. YANG (RFC 7950) is a data modeling language for NETCONF (RFC 6241), developed by the IETF NETMOD WG. This formatter provides support for Emacs yang-mode.


raxod502 (approved):
:sparkles: Thanks!

PR #284, opened 2024-01-30
Add support for dhall format

zzantares:
Adds a new formatter `dhall-format` used for [dhall](https://dhall-lang.org/).


raxod502:
Hmmm, if this formatter can automatically delete code comments, I am not sure we should enable it by default...

raxod502:
Is it supposed to be nondestructive and this is a bug? Or is it illegal to put comments in the place where there is one in your example?

zzantares:
good catch, it seems this is a bug that has been reported upstream: https://github.com/dhall-lang/dhall-lang/issues/1275.

raxod502:
OK, we can add support for the formatter, but not enable it by default, since it has this destructive behavior.

raxod502 (approved):
:sparkles:

PR #286, opened 2024-03-02
Update `apheleia-indent-lisp-buffer` to respect local variables

tpeacock19:
I ran into some issues using apheleia while contributing to a project
that works with `indent-tabs-mode` disabled. This led to this slight
modification of `apheleia-indent-lisp-buffer`

This allows for any local values for `indent-line-function`,
`lisp-indent-function` to be carried over when
formatting. Additionally, added `indent-tabs-mode` in order to
prevent adding tabs when unwanted.

If there is any additional documentation you would like me to provide,
let me know and I can update.
-------

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,15 @@
 All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog].
 
+## Unreleased
+### Enhancements
+### Formatters
+### Bugs fixed
+* `apheleia-indent-lisp-buffer` updated to apply local variables after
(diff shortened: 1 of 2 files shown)


raxod502:
Yeah, that makes sense. We should be enabling the major mode _before_ setting the local variables. I think this just needs a changelog entry, no need to update the README as the new behavior is what a user would intuitively expect to happen.

tpeacock19:
I put it under bugs fixed if that works

raxod502 (approved):
:sparkles: Thanks!

PR #288, opened 2024-03-04
Fix all clojure ts related major modes

dalugm:
After this pr

https://github.com/clojure-emacs/clojure-ts-mode/pull/36

The names of clojure-ts's major modes had changed


raxod502:
We should keep the old names for backwards compatibility, then, right?

dalugm:
> We should keep the old names for backwards compatibility, then, right?

I don't think so... In fact I highly doubt does anyone actually use it to program before v0.2.1. Because before v0.2.1, clojure-ts-mode can only highlight syntax, it lacked many features in clojure-mode and it could not work with cider (I believe 90% clojurists in Emacs use it). Even now (v0.2.2, the name changed happen in v0.2.1), it's just worked and still need improve.

But... I found some major modes like `clojure-ts-clojuredart-mode` are drived from `clojure-ts-mode`. We can keep the base mode only.

raxod502 (approved):
:sparkles: Ok!

PR #290, opened 2024-03-09
Prevent null issues when strings are unequal

elken:
In the case where s2 is larger than s1, this errors because the index is out of range.

A good testcase is the below

```lisp
(apheleia--align-point 
 "       <div class=\"left-[40rem] fixed inset-y-0 right-0 z-0 hidden lg:block xl:left-[50rem]\">\n  <svg\n"
 "<div class=\"left-[40rem] fixed inset-y-0 right-0 z-0 hidden lg:block xl:left-[50rem]\">\n <svg"
6)
```

If I've implemented the indexing wrong, do let me know but this seems to work just fine now for `mix` (the formatter that triggered this)


raxod502:
I'm going to fix up this unit testing framework a bit to be more usable, add some additional test cases for the string alignment function, and then merge. It might take a bit but I'll get it done.

elken:
Sounds good!

raxod502:
Okay, so I wrote some tests, and the test framework is very pretty now, but the tests are not passing... problem is, I'm pretty sure the tests are right and actually the string alignment algorithm has been wrong since I wrote it in 2019. Unfortunately I have no clue how it works anymore. So I will be figuring that out and fixing it, I guess.

I bet that will solve https://github.com/radian-software/apheleia/issues/2...

image

raxod502:
Aha, it WAS wrong!! And now it's fixed, I can't believe we've all (myself included) been living with the core algorithm of this package just being... wrong, for so long. Never again, I learned my lesson and will write tests.

raxod502 (approved):
:sparkles:

raxod502:
Sorry it took so long but I hope you'll agree the result is worth it!

PR #292, opened 2024-03-21
feat: handle bun projects in apheleia-npx

inkpot-monkey:
The apheleia-npx script now checks for bun projects and handles them correctly by using the installed bun bin exec the command.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@ The format is based on [Keep a Changelog].
 
 ## Unreleased
 ### Enhancements
+* `apheleia-npx` updated to handle bun projects ([#292])
 ### Formatters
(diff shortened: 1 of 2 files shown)


thomassdk:
Not sure why the formatter test is failing, as this change shouldn't have any impact on previous formatters.

Would it be useful for me to add in a test case here for bun?

raxod502:
image

Could be that a newer version of Prettier formats differently by default, breaking the test, or a bug in Prettier being able to find configuration files correctly. Probably the former. I can look into it in a bit.

A new test would be great although the test framework might need to be tweaked a bit to write a good test for this. That is also something I can look at in a bit.

Or you can have a go if you feel up to it earlier.

PR #293, opened 2024-03-22
Parse shebangs on Windows

ItsHoff:
Parses the shebang from formatter scripts on Windows, and inserts the binary in to the command. This fixes https://github.com/radian-software/apheleia/issues/285. 
Please let me know if there is still something that need more work, thanks.

One minor issue I noticed is that errors will now get logged to `*aphleia-bash-log*` instead of `*apheleia-npx-log*`.


raxod502 (approved):
:sparkles:

PR #301, opened 2024-05-01
Fix `apheleia-npx` in Yarn PnP projects

zeorin:
* `apheleia-npx` would use an incorrect path for the Yarn PnP ESM loader.
* `apheleia-npx` did not correctly guard against word splitting.
* `apheleia-npx` was sometimes not able to find formatters in a Yarn PnP project if there was also a node_modules folder at the root of the project. Unfortunately, many tools (including [Prettier](https://github.com/prettier/prettier/issues/13032)) will create a cache folder in `node_modules` even in Yarn PnP projects. The presence of any `node_modules` folders are irrelevant when a `.pnp.cjs` file is present.


zeorin:
cc @edslocomb & @raxod502, as you were involved in https://github.com/radian-software/apheleia/pull/200.

raxod502:
LGTM generally. Looks like upstream changes broke the tests. I'll fix, and also make sure we have test coverage that actually validates this change.

edslocomb (approved):
The path/filename bugfixes are great, it took me a while to figure out how the thing worked at all without them ;)

I had initially coded a preference for node_modules over pnp.js so as not to surprise anyone with changed behavior, but it does make a lot more sense to use pnp if present

Apologies to @zeorin for taking so long to look at this, notifications got lost in the shuffle

scripts/formatters/apheleia-npx
```diff
@@ -36,30 +36,32 @@ find_upwards() {
...
+            node="$(which node)"
             if [[ -f "${pnp_root}/.pnp.loader.mjs" ]]; then
-                loader_opt="--loader ${pnp_root}/.pnp.loader.mjs"
+                exec ${node} --require "${pnp_path}" \
+                    --loader "${pnp_root}/.pnp.loader.mjs" "${bin}" "${@:2}"
             fi
-            node=$(which node)
-            command="${node} --require ${pnp_path} ${loader_opt} ${bin} ${@:2}"
-            exec ${command}
+            exec ${node} --require "${pnp_path}" "${bin}" "${@:2}"
```
edslocomb, line 56: Are we doing two `exec`s now when .pnp.loader.mjs is present?

raxod502:
> Apologies to @zeorin for taking so long to look at this, notifications got lost in the shuffle

Also my fault because I was in the middle of working on this and it was still in progress but I have been busy with other things in my life and not attending to any of my GitHub notifications for a while, there are several dozen threads that have been waiting. This is still very much on my list if Ed doesn't beat me to it.

scripts/formatters/apheleia-npx
raxod502: Nah, `exec` is a terminal command. It completely replaces the current process with that of the invoked command - so, nothing after it in the script will execute. https://en.wikipedia.org/wiki/Exec_(system_call)

raxod502:
Heyo, it turns out that I had got this pretty close to working already! The latest few changes seem to make things fully successful and now we have a proper test framework that supports ensuring that features like this will not break in the future.

raxod502 (approved):

PR #316, opened 2024-08-26
Update rubocop command

uvera:
`--auto-correct` got deprecated in https://github.com/rubocop/rubocop/commit/aabdf0acf5ed729daf20506475d5629bfffb1890

### DIFF
--- a/apheleia-formatters.el
+++ b/apheleia-formatters.el
@@ -158,7 +158,7 @@
     (python3-json
      . ("python3" "-m" "json.tool"
         (apheleia-formatters-indent "--tab" "--indent")))
-    (rubocop . ("rubocop" "--stdin" filepath "--autocorrect"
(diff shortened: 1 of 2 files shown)


apheleia-formatters.el
```diff
@@ -158,7 +158,7 @@
     (python3-json
      . ("python3" "-m" "json.tool"
         (apheleia-formatters-indent "--tab" "--indent")))
-    (rubocop . ("rubocop" "--stdin" filepath "--auto-correct"
+    (rubocop . ("rubocop" "--stdin" filepath "--autocorrect"
```
edslocomb, line 161: This will break old or maintenance-mode projects using prior versions of rubocop; can we just use the "-a" alias?

edslocomb (approved):

PR #319, opened 2024-09-18
mix needs to use --stdin-filename to support non-elixir files like `.heex`

LucianoLaratelli:
Hey, hope all is well. 

I was seeing errors formatting `.heex` files from `elixir-mode` in `*apheleia-apheleia-from-project-root-log*`:
```
$ apheleia-from-project-root .formatter.exs mix format -

mix format failed for stdin
** (SyntaxError) invalid syntax found on stdin.exs:1:1:
    error: syntax error before: '<'
    │
  1 │ <html lang="en" class="[scrollbar-gutter:stable]">
    │ ^
    │
    └─ stdin.exs:1:1
    (elixir 1.17.2) lib/code.ex:999: Code.format_string!/2
    (mix 1.17.2) lib/mix/tasks/format.ex:649: Mix.Tasks.Format.elixir_format/2
    (mix 1.17.2) lib/mix/tasks/format.ex:668: Mix.Tasks.Format.format_file/2
    (elixir 1.17.2) lib/task/supervised.ex:101: Task.Supervised.invoke_mfa/2
    (elixir 1.17.2) lib/task/supervised.ex:36: Task.Supervised.reply/4
```
Looking through the help page for `mix format`, I saw:
> `--stdin-filename' - path to the file being formatted on stdin. This is useful if you are using plugins to support custom filetypes such as .heex. Without passing this flag, it is assumed that the code being passed via stdin is valid Elixir code. Defaults to "stdin.exs".

This makes that change.

### DIFF
--- a/apheleia-formatters.el
+++ b/apheleia-formatters.el
@@ -90,8 +90,8 @@
     (lisp-indent . apheleia-indent-lisp-buffer)
     (ktlint . ("ktlint" "--log-level=none" "--stdin" "-F" "-"))
     (latexindent . ("latexindent" "--logfile=/dev/null"))
-    (mix-format . ("apheleia-from-project-root"
-                   ".formatter.exs" "mix" "format" "-"))
+    (mix-format . ("apheleia-from-project-root" ".formatter.exs"
+                   "mix" "format" "--stdin-filename" filepath "-"))
     (nixfmt . ("nixfmt"))
     (ocamlformat . ("ocamlformat" "-" "--name" filepath
                     "--enable-outside-detected-project"))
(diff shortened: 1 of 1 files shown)


raxod502:
Looks like CI is failing thusly, is this option not widely available yet?

image

LucianoLaratelli:
Ah, sorry about that. 

It looks like the earliest version that had the option available was [1.14.0](https://hexdocs.pm/mix/1.14.0-rc.0/Mix.Tasks.Format.html#module-task-specific-options). The first release candidate for that version was cut in August 2022, so just a little more than two years ago. The current version is 1.17.3.

raxod502:
The dilemma I have is that Apheleia doesn't really know which version of a tool it's executing.

When I get some time, I will think on this. I think a framework can be introduced to allow defining how Apheleia will determine the installed version of a tool and cache the result, then use that to condition on what arguments it will pass.

Maybe we can get away with something as simple as implementing it in a wrapper script like `apheleia-npx`.

raxod502:
Check the implementation in https://github.com/radian-software/apheleia/pull/322

LucianoLaratelli:
Ah, very nice. Thanks for doing that :)

PR #331, opened 2024-11-12
Add Runic.jl as formatter for julia

jpthiele:
This will add (Runic.jl)[https://github.com/fredrikekre/Runic.jl] as a formatter for (Julia)[https://julialang.org/].
Instead of a `PATH` it is sufficient to have an alias as described in the Readme of Runic.


test/formatters/installers/runic.bash
```diff
@@ -0,0 +1,5 @@
+# install recent julia through JuliaUp
+curl -fsSL https://install.julialang.org | sh -s -- --yes
+ln -s /root/.juliaup/bin/julia /usr/bin/julia
+# install Runic.jl in @runic shared project
+julia --project=@runic -e 'using Pkg; Pkg.add(Runic)'
```
fredrikekre, line 5:
```suggestion
julia --project=@runic -e 'using Pkg; Pkg.add("Runic")'
```

raxod502:
> Since I have podman instead of docker I can't test everything locally

Fixed that for you in https://github.com/radian-software/apheleia/pull/343

PR #339, opened 2024-12-25
feat: add biome formatter

danielpza:
See https://biomejs.dev/

Alternative formatter to `prettier`

closes https://github.com/radian-software/apheleia/issues/323


apheleia-formatters.el
```diff
@@ -1177,7 +1178,7 @@ For more implementation detail, see
     (setq-local indent-line-function
                 (buffer-local-value 'indent-line-function buffer))
     (setq-local lisp-indent-function
-		(buffer-local-value 'lisp-indent-function buffer))
+                (buffer-local-value 'lisp-indent-function buffer))
```
danielpza, line 1181: Adjusted the indentation in this line to be consistent with the whole file. I can revert it if undesired.

apheleia-formatters.el
```diff
@@ -35,6 +35,7 @@
               "-"))
     (brittany . ("brittany"))
     (buildifier . ("buildifier"))
+    (biome . ("apheleia-npx" "biome" "format" "--stdin-file-path" filepath))
```
martinjlowm, line 38: FYI - `format` does not sort imports (for some strange reason), however applying both `check` and `--fix` to the command then it does.

apheleia-formatters.el
raxod502: Oh oof. Thanks for fixing that.

raxod502 (approved):
Thank you, this is great, sorry for the delay!

PR #341, opened 2025-01-18
Resolve formatter executable commands prior to calling them

tylerjl:
Hi! Thanks for your work on apheleia, it's great.

An issue I've run into recently is that sometimes my remote executables aren't immediately in-`$PATH`: I need to resolve them via `executable-find` - which does find them - before invoking the process. This should be safe, given that the step is performed at a prior step. I'm overwriting the `command` variable since I infer that the executable _name_ is being reused later when you're matching a callback, I think?

I mentioned that `executable-find` happens earlier in the call chain so I could maybe improve this by removing that and only resolving the executable once, but I wanted to check in with you to determine whether the overall problem and solution even make sense to somebody else.


raxod502:
See also: https://github.com/radian-software/apheleia/issues/347

raxod502:
I think this is a workaround to a problem that is better solved elsewhere, but I think it is okay to do this to fix the immediate problem rather than waiting to fix everything.

raxod502 (approved):

tylerjl:
Thank you @raxod502, I appreciate it :pray:

PR #345, opened 2025-01-24
Configuration examples

### LINKED ISSUE #344: Show examples of (1) 'use-package' for configuration, (2) enabling a less popular mode, and (3) defining a formatter
### Documentation

This is a great package and works well with just this in `user-init-file`
```elisp
  (apheleia-global-mode)
```

But when I tried to use a less popular mode -- `markdown-mode` it took me hours to figure out how to configure it as elegantly as the packaga allows.  (I spent lots of time trying ChatGPT-generated elisp that worked, but was convolutionarily complexicated).  After I got it working _and_ minimal, I want to share what I learned so that others can use this package to it's best effect instead of being stopped.

Even though the documentation was correct (in hind sight), it wasn't as helpful (in fore sight) as I would have hoped.  It would be better to have an working _example_ (or 3) than a _definition_ of all the pieces.

### Formatting

Ironically, I was trying to enable `markdown-mode`, so that once I had enabled auto-format for my own `*.md` files, emacs exposed that `README.md` was not compliant with the package's own `prettier-markdown` formatter.

MichaelRWolf:
- Add examples to `README.md`
- Allow `prettify-markdown` to do Apheleia-magic to itself.

Resolves #344

### DIFF
--- a/README.md
+++ b/README.md
@@ -95,7 +95,8 @@ following code will enable Apheleia globally to use the popular
 formatters in `apheleia-formatters` if associated with popular modes
 in `apheleia-mode-alist`.
 
-* `apheleia-formatters`: Alist mapping names of formatters (symbols
+- `apheleia-formatters`: Alist mapping names of formatters (symbols
+
 ```elisp
 (use-package apheleia
   :ensure t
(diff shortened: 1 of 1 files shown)


raxod502:
Hmmm. I like the added documentation, but is there a reason that the formatting from prettify-markdown is better than the previous version? It just seems different.

PR #352, opened 2025-03-04
Specify file type for buildifier explicitly

avdv:
When formatting Bazel files, buildifier needs to know the concrete file
type which it usually deduces from the file name which is impossible
when operating on stdin.

For example, the order of load statements in a WORKSPACE file is
significant and re-ordering might break it.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,10 @@
 All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog].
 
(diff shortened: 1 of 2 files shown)


raxod502 (approved):

raxod502:
If there's an option in this tool to supply the filename as a command-line argument, we can do that instead to avoid having to re-implement the inference ourselves. Merging this for now as an incremental improvement.

PR #353, opened 2025-03-08
Fix patches moving a line to the top of the file.

### LINKED ISSUE #299: RCS patch wrongly applied results in incorrect file
Consider the following file in a Javascript project with Prettier and the plugin https://github.com/IanVS/prettier-plugin-sort-imports (which makes Prettier automatically sort imports):
```js
import {
    firstLongImport,
    secondLongImport,
    thirdLongImport,
} from "longDependency";
import { fourthImport } from "dependency";
```
Formatting it in Emacs with Apheleia results in the following invalid code:
```js
import {
import { fourthImport } from "dependency";
    firstLongImport,
    secondLongImport,
    thirdLongImport,
} from "longDependency";
```
(the expected behavior is that `fourthImport` is moved to the first line).

Running `apheleia-npx` from the terminal with the correct arguments correctly formats the file, so the issue seems to be in Apheleia, not in the formatter.
Here is the RCS diff I get, running `diff` in the terminal with the same arguments as Apheleia):
```
a0 1
import { fourthImport } from "dependency";
d6 1
```

Could it be that the `a0` line isn’t handled correctly?

guillaumebrunerie:
RCS patches containing a 'a0' instruction are not handled properly.
This pull request fixes the handling of 'a0' instructions and adds an integration test that fails otherwise.
Fixes #299.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,10 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog].
 
 ## Unreleased
+### Bugs fixed
+* A formatter that moves a line to the top of the file would sometimes
+  place it as the second line instead ([#299]).
+
 ### Formatters
 * Format Bazel files according to their type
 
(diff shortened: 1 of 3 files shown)


raxod502 (approved):
Beautiful! Thank you! I'm glad the framework I started was easy enough to extend.

You mentioned that you had trouble running the tests on macOS - if you feel like it, go ahead and file another issue with the error message from that, I suspect it's easy to fix.

PR #356, opened 2025-03-30
Pass --text arg to diff

jkxyz:
For some reason, GNU diff on Windows (from the MinGW project) fails to diff the file contents when passed over stdin unless the `--text` argument is passed. This argument forces diff to treat the inputs as text, and not binary.

To reproduce this error with Bash on Windows:

```
$ echo foo > foo
$ echo bar > bar
$ cat bar | diff.exe foo -
diff.exe: -: Invalid argument
$ cat bar | diff.exe --text foo -
1c1
< foo
---
> bar
```

It seems like this flag is harmless in this case, since Apheleia will always be used for diffing text files, and never binary files.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,8 @@ The format is based on [Keep a Changelog].
 ### Bugs fixed
 * A formatter that moves a line to the top of the file would sometimes
   place it as the second line instead ([#299]).
+* Fix invoking the diff command on Windows by always passing the --text
+flag
 
 ### Formatters
 * Format Bazel files according to their type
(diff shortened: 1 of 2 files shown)


DinoChiesa:
Strange, because Apheleia on windows works for me.  I am using the diff utility that is packaged with the Git program, found in `c:\Program Files\Git\usr\bin\diff.exe` 

```
> diff.exe --version
diff (GNU diffutils) 3.10
Copyright (C) 2023 Free Software Foundation, Inc.
License GPLv3+: GNU GPL version 3 or later <https://gnu.org/licenses/gpl.html>.
This is free software: you are free to change and redistribute it.
There is NO WARRANTY, to the extent permitted by law.

Written by Paul Eggert, Mike Haertel, David Hayes,
Richard Stallman, and Len Tower.
```

But yes, `--text` ought to be harmless in any case.

raxod502 (approved):

PR #357, opened 2025-03-30
Handle Windows and remote buffers using default formatters

jkxyz:
Various fixes to ensure that the default formatters work over Tramp and on Windows:

1. Replace the apheleia-npx command with npx when the buffer is remote or running on Windows. This uses the built-in npx handling to lookup the binary, so npx never gets called. Since the apheleia-npx script isn't available on remote hosts, it makes sense to remove this. The same is done for Windows to capture the case of using WSL, in which case calling bash will enter WSL and so the script cannot be accessed using the Windows path.

2. Use local file name when returning binary path.

3. Don't use executable-find if the binary is already an absolute path.


apheleia-formatters.el
```diff
@@ -968,6 +971,13 @@ cmd is to be run."
       (setf (apheleia-formatter--name context) name)
       (setf (apheleia-formatter--stdin context) stdin)
       (setf (apheleia-formatter--remote context) remote)
+
+      ;; Just use npx when buffer is remote or we're running on Windows,
+      ;; since in these cases bash won't be able to find the script
+      (when (and (member "apheleia-npx" command)
+                 (or run-on-remote (eq system-type 'windows-nt)))
+        (setq command (cons 'npx (remove "apheleia-npx" command))))
```
raxod502, line 979: This will be an issue for any of the scripts in the Apheleia scripts directory, just just the npx one. What would you think of having Apheleia copy the script to the remote host before first execution, if anything `apheleia`-prefixed is used in the command?

PR #358, opened 2025-04-02
fix: use proper path separator on Windows

DinoChiesa:
Fix the use of ":" as a separator for path elements. This is inappropriate for Windows.
There's a variable we can use instead: `path-separator`. 

I don't know exactly what breaks when the wrong separator is used. I do know that it's common for Windows paths to include ":" because that is the separator between the drive letter and the drive-local path.


DinoChiesa:
regarding the 79-char limit, ironically, I don't have an elisp formatter that re-formats the 79 characters limit.  The default formatter I have is `lisp-indent` .  is there a different setting I need to apply to that thing?

raxod502:
I just do it manually, I'm not aware of any re-flowing Elisp formatters.

We should add a changelog entry though, if one is needed.

PR #361, opened 2025-05-16
feat: add nomad formatter

Fuco1:
This is a more specific formatter than `hclfmt` for nomad files.


raxod502 (approved):
:heart: :sparkles:

PR #363, opened 2025-05-17
Fix moving point

tmurph:
When replacing text, calculate the cursor offset relative to the start of the replacement, not the end of the text-to-be-added.

* test/integration/apheleia-it.el (supports-inserting-whitespace): New test case
* apheleia-rcs.el (apheleia--apply-rcs-patch): Go to start of deletion text before calculating offsets


raxod502 (approved):
:heart: :sparkles:

PR #370, opened 2025-07-18
add yaml-ts-mode to aphelia-formatters-indent

WJCFerguson:
Trivial change - other modes include `-ts-mode` variants, but not yaml.

This fix is confirmed to work for me, taking the indent level from `yaml-indent-offset`.


raxod502 (approved):
Thanks!

PR #380, opened 2025-12-09
Pass --force-exclusion to rubocop

ibrahima:
This make rubocop obey exclusions from config files. Otherwise, even if a file is excluded from in .rubocop.yml, apheleia will still format it.

See https://docs.rubocop.org/rubocop/usage/basic_usage.html#command-line-flags


raxod502:
Looking at the test failure, is it because there is a new version of rubocop and it changed the default formatting heuristics?

PR #385, opened 2026-01-17
Remove --no-doc argument from yq-yaml formatter 

akurth:
The --no-doc argument removes existing document separators (---) from yaml files. It is perfectly valid and common practice to use multiple documents within a single file, e.g. in the kubernetes universe. Formatting such files with --no-doc breaks their validity.

See also the prettier-yaml test, which preserves the document separator, too.


akurth:
Sorry, I overlooked that old commit c8e1502 (no idea why it’s showing up here, since it has already been merged years ago).

I’ll close this PR and create a clean new one.

PR #386, opened 2026-01-18
Remove --no-doc argument from yq-yaml formatter

akurth:
The --no-doc argument removes existing document separators (---) from yaml files. It is perfectly valid and common practice to use multiple documents within a single file, e.g. in the kubernetes universe. Formatting such files with --no-doc breaks their validity.

See also the prettier-yaml test, which preserves the document separator, too.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,8 @@ The format is based on [Keep a Changelog].
 
 ## Unreleased
 ### Formatters
+* Removed `--no-doc` argument from yq-yaml to avoid breaking
+  multi-document yaml files ([#386]).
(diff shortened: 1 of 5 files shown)


raxod502:
Can we also add a test checking that, with the new change, existing documents _without_ document separators don't get separators automatically added?

akurth:
Sure, just pushed the additional test.

raxod502:
Cool, and then changelog entry?

akurth:
I added the CHANGELOG line, squashed everything into one commit and force-pushed.

raxod502 (approved):
Awesome! Thanks!

PR #388, opened 2026-03-01
fix: support for gdiff in UNIX-like systems

yukiteruamano:
Hi!

In OpenBSD and  FreeBSD, `diff` don't have support for `--rcs`. In these cases, we need to use `gdiff` or `GNU diff` for this support. 

This change, fix the use of `diff `or `gdiff` for these OSes and make apheleia work.

### DIFF
--- a/apheleia-formatters.el
+++ b/apheleia-formatters.el
@@ -938,7 +938,10 @@ See `apheleia--run-formatters' for a description of REMOTE."
 
     (let ((ctx (apheleia-formatter--context)))
       (setf (apheleia-formatter--name ctx) nil ; Skip logging on failure
-            (apheleia-formatter--arg1 ctx) "diff"
+            (apheleia-formatter--arg1 ctx)
+            ;; Search gdiff in PATH, else use diff, useful for BSD
+            (if (executable-find "gdiff") "gdiff"
+              "diff")
             (apheleia-formatter--argv ctx) `("--rcs"
                                              "--strip-trailing-cr"
                                              "--text"
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,14 @@
 All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog].
 
+## Unreleased
+### Enhancements
+
+* Better support for BSD OSes using `gdiff` for RCS formatting, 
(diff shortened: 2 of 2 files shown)


apheleia-formatters.el
```diff
@@ -534,8 +534,8 @@ and then write the formatted output back to the remote machine. Note some
 features of `apheleia' (such as `file' in `apheleia-formatters') is not
 compatible with this option and formatters relying on them will crash."
   :type '(choice (const :tag "Run the formatter on the local machine" local)
-                 (const :tag "Run the formatter on the remote machine" remote)
-                 (const :tag "Disable formatting for remote buffers" cancel))
+          (const :tag "Run the formatter on the remote machine" remote)
+          (const :tag "Disable formatting for remote buffers" cancel))
```
raxod502: Accidental indentation change?

apheleia-formatters.el
```diff
@@ -938,7 +938,10 @@ See `apheleia--run-formatters' for a description of REMOTE."
 
     (let ((ctx (apheleia-formatter--context)))
       (setf (apheleia-formatter--name ctx) nil ; Skip logging on failure
-            (apheleia-formatter--arg1 ctx) "diff"
+            (apheleia-formatter--arg1 ctx)
+            (if (memq system-type '(berkeley-unix freebsd openbsd))
+                "gdiff"
+              "diff")
```
raxod502, line 944: For a more robust check, what about checking if `gdiff` is on the PATH, and using that if found, else using `diff`?

apheleia-formatters.el
yukiteruamano: Yeah, I fix this in next commit

apheleia-formatters.el
yukiteruamano: Good, in next commit I try this approach, thanks!

yukiteruamano: Done

raxod502:
Looks good, mind adding a changelog entry?

yukiteruamano:
> Looks good, mind adding a changelog entry?

Not problem. 

Regarding the changes, I've been using them in my version of Apheleia and Doom Emacs without any problems.

raxod502 (approved):

PR #389, opened 2026-03-11
OCaml: Add neocaml modes for ocamlformat

mattiasdrp:
Add neocaml-mode and neocaml-interface-mode to the list of modes that can be formatted with ocamlformat


raxod502 (approved):
Thanks!

PR #391, opened 2026-04-14
Add OpenTofu, use by default for Terraform

radon-at-beeper:
Seems fairly straightforward to me, Terraform is no longer open-source, OpenTofu is the compatible open-source successor, similar to the Redis/Valkey situation.

Of course, if someone still prefers to use Terraform instead, they can easily customize using:

```tf
(setf (alist-get 'terraform-mode apheleia-mode-alist) 'terraform)
```

We don't have a mechanism for allowing multiple formatters to be registered for the same major mode, where later ones could be used if earlier ones are not installed. That might be a nice addition.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,11 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog].
 
 ## Unreleased
+### Changes
+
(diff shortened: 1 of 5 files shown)


raxod502 (approved):
Great PR

Image

PR #395, opened 2026-05-15
Add sqlfluff formatter for SQL

ljohaansson:
Add sqlfluff as an opt-in formatter.

Tested with `make fmt-build fmt-docker FORMATTERS=sqlfluff CMD='make fmt-test'`.


raxod502 (approved):
Thanks!

PR #402, opened 2026-06-02
fix(shfmt): do not pass -ln or --indent, to respect editorconfig

scop:
This might be a bit controversial, but without this, as it stands, there's no way to get apheleia uses of shfmt to honor projects editorconfigs (`-ln` is passed unconditionally). FWIW in my opinion shfmt's behavior is not great there, better would be to override only editorconfig settings by ones set on the cmdline, not all of them even if anything is passed. But it seems it's intentional in shfmt.

Ref https://github.com/mvdan/sh/blob/master/cmd/shfmt/shfmt.1.scd

> If any parser or printer flags are given to the tool, no EditorConfig formatting options will be used.

Users are better off using editorconfig to configure shfmt, so the configuration applies to all kinds of shfmt uses.

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -4,6 +4,7 @@ on:
     branches:
       - main
   pull_request: {}
+permissions: {}
 jobs:
   lint:
     runs-on: ubuntu-latest
@@ -21,5 +22,6 @@ jobs:
       - name: Run linters
         env:
           VERSION: ${{ matrix.emacs_version }}
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
         run: >-
           make docker CMD="make unit integration lint lint-changelog"
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,10 +20,16 @@ The format is based on [Keep a Changelog].
 * Recognize csharp, dockerfile and php ts-modes, apply same settings as
   for the corresponding non-ts modes ([#396]).
 
+### Bugs fixed
+
+* Do not pass any parser or printer options to shfmt. Doing so causes
+  .editorconfig configuration that might be in effect to be ignored
+  altogether ([#401]).
+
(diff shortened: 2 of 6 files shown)


scop:
Sample project currently affected badly by this: https://github.com/scop/bash-completion

We have very specific config for shfmt in .editorconfig, but because apheleia passes -ln to shfmt, none of that gets applied. Similarly, passing --indent would result in the same.

raxod502:
I don't think it's quite appropriate to put in the bugfixes section, so I rewrote the changelog entry to put this simply under "changes". But, I see your concern. We can make this change, and if anyone complains about the new behavior, we can perhaps make it a configurable option which way you want Apheleia to work.

Maybe you can also file an issue upstream with shfmt, seeing if they would consider making it possible to apply command-line options and just have them override the relevant .editorconfig settings they override, instead of demolishing everything? Seems like a reasonable change to me.

scop:
Thank you for considering this!

I dropped some notes to an existing upstream issue about the same thing, https://github.com/mvdan/sh/issues/975#issuecomment-4637894724

Having thought of this for a bit, I'm starting to lean towards shfmt's behavior being arguably fine, or at least not being at all sure that merging settings from editorconfig and command line would be The Right Thing instead.

raxod502:
Well, the issue in my mind is not so much that shfmt ignores some editorconfig options when command-line options are provided. Because as you observe in that comment on the other issue, one can configure Emacs to read from editorconfig and then Emacs could pass the right options via Apheleia. That seems fine.

The problem is that there are some options that can be configured for shfmt in .editorconfig that are not necessarily going to be supported by the Emacs editorconfig integration. Because what variables would they be setting? They would be shfmt-specific, and the editorconfig package can't have explicit support for every individual language, there rather needs to be a generic mechanism for per-language settings that could be passed along in its whole by Apheleia. But I don't see something like that.

But, whatever.

PR #409, opened 2026-06-18
Fix window scroll jump after line-neutral RCS patches

jarreds:
## Summary

Fix a scroll jump that can happen when applying an RCS patch changes
text around point without changing the point's vertical line position.

Apheleia restores scroll position by comparing the number of lines
between `window-start` and point before and after applying the patch.
For line-neutral replacements, such as removing leading whitespace
from the line containing point, Emacs can adjust `window-start` while
the replacement is applied. In that case, Apheleia may scroll even
though point did not move vertically relative to the original window
start.

This stores the original `window-start` marker and line distance. If
point is still the same number of lines from the original window start
after the patch, restore that window start directly; otherwise keep
the existing scroll adjustment behavior.

## Test

Added a unit test covering a one-line replacement where point moves
from an indented column to beginning of line.

Before the fix, the new test fails with `window-start` moving from
line 11 to line 9. With the fix, `make unit` passes.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -38,6 +38,12 @@ The format is based on [Keep a Changelog].
 * [taplo](https://taplo.tamasfe.dev/) is now used by default for TOML
   also in `conf-toml-mode` ([#407]).
 
+### Bugs fixed
+
+* Apheleia no longer scrolls the window after formatting when a
+  line-neutral change only moves point to the beginning of its line
+  ([#409]).
+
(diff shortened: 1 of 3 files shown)


jarreds:
How I noticed this: in a saved buffer, put a bunch of spaces before some characters on a line, save, then after formatting the buffer shifts.

raxod502 (approved):
Thank you!

PR #412, opened 2026-09-02
Apply Biome assist actions during formatting

jkxyz:
Biome assist actions such as `useSortedPackageJson` do not run under `biome format`. Use `biome check --write` with linting disabled so configured assists run without applying lint fixes.

Tested with `make fmt-test FORMATTERS=biome` and `make fmt-lint`.


jkxyz:
It looks like the testing workflow was originally failing because the Docker image started picking up Emacs 31, which emits new warnings about obsolete macros that the output parsing wasn't able to handle. The second commit includes a fix for that.

raxod502:
Thank you for the help! That all makes sense to me. Smart thinking with using sed on stderr without interleaving it.

raxod502 (approved):

PR #414, opened 2026-09-15
feat(compability): replace `when-let` and `if-let` with its `*` count…

padulkemid:
this will replace deprecated functions and will stop the warning in startup.


raxod502 (approved):
Thanks!

